PR #8, opened 2018-01-25
Core configuration parser

wtrocki:
### Motivation

Core interface should provide way to manage services. 
Proposed architecture reflects completely architecture proposed for Android:

- Core is available as singleton for general usage
- Separate instances of core can be created if neede
- Configuration is hold in memory. Services query core for configuration
- Core contains http and logger interface.  

## Progress
- [x] Investigate and build and configuration parsing for Core module
- [x] Provide lightweight and configurable logging solution that can be disabled by users.
- [x] Refactor implementation to properly handle errors (guards etc.)
- [x] Add network layer
- [x] Write unit tests for implementations

## Sample app

Sample application for research/investigation purposes


wtrocki:
@danielpassos Fully agree. We have plenty of time for this and this work should be really separate PR to simplify review process. Thank you for review.

david-martin:
@wtrocki @danielpassos I think it would be good to focus on the follow up work now.
@wtrocki can you create the issue(s)

wtrocki:
Issue created: https://issues.jboss.org/browse/AGIOS-646

PR #11, opened 2018-02-02
Network request wrapper

wtrocki:
### Motivation

Remove AeroGearHttp as we still do not know how this will be maintained.
Use alamofire for dependency management.
Provide simple wrapper for http api. 
Wrapper was simplified as Alamofire provides many extensible options and we are not sure what may be required at the moment. 
We could extend that over the time basing on what our SDK's will need.

#### Notes

It was not possible to keep similar api for android as Android implementation it's not using lambdas. 
As lambdas are TODO for Android it's best to align Android implementation to IOS.

### DIFF
--- a/modules/core/http/AgsHttp.swift
+++ b/modules/core/http/AgsHttp.swift
@@ -1,4 +1,3 @@
-import AeroGearHttp
 import Foundation
 
 public protocol Request {
@@ -9,22 +8,15 @@ public protocol Request {
  */
 public class AgsHttp {
 
-    let defaultHttp = Http()
+    let defaultHttp = AgsHttpRequest()
 
     public init() {
     }
 
     /**
-     * Return new Http instance for specific service
-     */
-    public func getHttp(service: MobileService) -> Http {
-        return Http(baseURL: service.config?.uri)
-    }
-
-    /**
      * Return shared Http instance
      */
-    public func getHttp() -> Http {
+    public func getHttp() -> AgsHttpRequest {
         return defaultHttp
     }
 }
--- a/modules/core/http/AgsHttp.swift
+++ b/modules/core/http/AgsHttp.swift
@@ -1,8 +1,5 @@
 import Foundation
 
-public protocol Request {
-}
-
 /**
  * Wrapper class used for network requests
  */
--- a/modules/core/http/AgsHttpRequest.swift
+++ b/modules/core/http/AgsHttpRequest.swift
@@ -1,0 +1,56 @@
+import Alamofire
+import Foundation
+
+/**
+ * This is a implementation of HttpRequest based on AlamoFire
+ * Implementation is designed to work with Json payload
+ */
+public class AgsHttpRequest {
+
+    public func get(_ url: String, params: [String: AnyObject]? = [:], headers: [String: String]? = [:],
+                    _ handler : @escaping (Any?, Error?) -> Void) {
+        Alamofire.request(url, parameters: params, headers: headers).responseJSON { (responseObject) -> Void in
+            if responseObject.result.isSuccess {
+                handler(responseObject.result.value, nil)
+            }
+            if responseObject.result.isFailure {
+                handler(nil, responseObject.result.error)
+            }
+        }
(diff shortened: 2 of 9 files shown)


modules/core/http/AgsHttp.swift
```diff
@@ -1,4 +1,3 @@
-import AeroGearHttp
 import Foundation
 
 public protocol Request {
```
danielpassos: Is this used somewhere?

modules/core/http/AgsHttpRequest.swift
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * This is a implementation of HttpRequest based on AlamoFire
+ * Implementation is designed to work with Json payload
+ */
+public class AgsHttpRequest {
+
+    public func get(_ url: String, params: [String: AnyObject]? = [:], headers: [String: String]? = [:],
+                    _ handler : @escaping (Any?, Error?) -> Void) {
+        Alamofire.request(url, parameters: params, headers: headers).responseJSON { (responseObject) -> Void in
+            if responseObject.result.isSuccess {
```
pb82: Is `responseObject` always guaranteed to be defined?

modules/core/http/AgsHttpRequest.swift
```diff
@@ -0,0 +1,56 @@
...
+                handler(nil, responseObject.result.error)
+            }
+        }
+    }
+
+    public func put(_ url: String, body: [String: AnyObject]? = [:], headers: [String: String]? = [:], _ handler: @escaping (Any?, Error?) -> Void) {
+        Alamofire.request(url, method: .put, parameters: body, encoding: JSONEncoding.default, headers: headers).responseJSON { (responseObject) -> Void in
+
+            if responseObject.result.isSuccess {
+                handler(responseObject.result.value, nil)
```
pb82: Is the convention in Swift to have the error as the last argument in the callback (because in JS it's the other way)?

modules/core/http/AgsHttpRequest.swift
wtrocki: Code will not compile otherwise:
https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/OptionalChaining.html

I'm using here AlamoFire code snippets wrapping them into interface. 
There is no changes between AlamoFire usage examples.

modules/core/http/AgsHttpRequest.swift
```diff
@@ -0,0 +1,56 @@
...
+            }
+            if responseObject.result.isFailure {
+                handler(nil, responseObject.result.error)
+            }
+        }
+    }
+
+    public func post(_ url: String, body: [String: AnyObject]? = [:], headers: [String: String]? = [:], _ handler: @escaping (Any?, Error?) -> Void) {
+        Alamofire.request(url, method: .post, parameters: body, encoding: JSONEncoding.default, headers: headers).responseJSON { (responseObject) -> Void in
+            if responseObject.result.isSuccess {
```
pb82: To avoid repeating this three times, can't we define one generic handler that is used? E.g.

```
Alamofire.request(url, method: .post, parameters: body, encoding: JSONEncoding.default, headers: headers).responseJSON(handleResponse)
```

modules/core/http/AgsHttpRequest.swift
wtrocki: Not sure about conventions, but most of the callbacks I know have similar aproach:
https://developer.apple.com/documentation/foundation/urlsession/1410330-datatask

pb82: Ok, if Apple uses that order then we also should

modules/core/http/AgsHttpRequest.swift
wtrocki: I think it's essential to have this separate for maintenance purposes. 
Closures are typical to be defined inline. 
There is not much actual business code in this so DRY it's not violated as much :)

danielpassos:
I will allow it

PR #12, opened 2018-02-06
Add class name to generator

wtrocki:
### Motivation

Classes need to start with capital letter. 
Allow to specify SDK name using small letter but also provide class name as capital.


wtrocki:
We may move from bash script to some cmd at some point. For the moment that's all it's needed.

danielpassos:
👍

PR #13, opened 2018-02-06
Metrics sdk

wtrocki:
## Motivation

Initial version of Metrics Swift SDK

## Progress

- [x] Initial SDK setup
- [x] Client ID
- [x] mobile service configuration (URL)
- [x] Investigate SDK version in source code
- [x] General refactoring and styles
- [x] Convert to new mobile-config.json format

### Handling network availability/state (retrying when failed)

For this I think we may split metrics collection and metrics upload into two processes.
Metrics could be processed and stored locally. 
Metrics service should send metrics once available and try to retry when no internet connection etc.

### Battery efficiency

Sending versions for every application launch seems to be redundant.
Ideally we would like to track/send updates only when value will change.

### Retrieving Version in IOS

IOS do not have mechanism to know library version as this is detached from cocoapods


modules/metrics/metrics/SdkVersion.swift
```diff
@@ -0,0 +1,76 @@
...
+        bundleId = Bundle.main.bundleIdentifier ?? "Unknown";
+    }
+ 
+    fileprivate func setupUri() {
+        // TODO
+        //var config = core.getConfiguration("metrics");
+    }
+    
+    fileprivate func initId() {
+        if let uuid = UIDevice.current.identifierForVendor?.uuidString {
```
wei-lee: not sure if `identifierForVendor` fits the purpose here. 

According to this from apple:

> The value of this property is the same for apps that come from the same vendor running on the same device.

I believe it will return the same value for the same developer. In another word, if I have 3 different apps installed on the same device, it will return the same value for all 3 apps. Not sure this is what you are looking for?

example/AeroGearSdkExample/AppDelegate.swift
```diff
@@ -21,12 +23,15 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
     func applicationDidEnterBackground(_: UIApplication) {
+        metrics.background();
     }
 
     func applicationWillEnterForeground(_: UIApplication) {
+    
     }
 
     func applicationDidBecomeActive(_: UIApplication) {
+        metrics.foreground();
```
wei-lee: the method name is a bit confusing. In metrics it is called `foreground`, but you are calling it from `applicationDidBecomeActive`, while there is also a `applicationWillEnterForeground` method here?

wtrocki: This is temporary and will be replaced with proper metrics implementation. 
Our metrics can hook into application lifecycle and other specific events.

Going to rename method and remove background method as it's not used.

modules/metrics/metrics/SdkVersion.swift
wtrocki: I checked couple options:
- custom libraries on github
- [advertisingidentifier](https://developer.apple.com/documentation/adsupport/asidentifiermanager/1614151-advertisingidentifier)
- `identifierForVendor` 

Was thinking that identifierForVendor will be the most appropriate option, but in fact I will go with the code that is available in else statement. 

I see that metrics and analytics for most of the cases are tied to user like here:
https://developers.google.com/analytics/devguides/collection/ios/v3/user-id

wei-lee: yeah, I think generating the device id and save it locally is enough for this use case here.

modules/metrics/metrics/SdkVersion.swift
```diff
@@ -0,0 +1,77 @@
...
+        setupUri()
+        setupBundleId()
+    }
+
+    fileprivate func setupBundleId() {
+        bundleId = Bundle.main.bundleIdentifier ?? "Unknown"
+    }
+
+    fileprivate func setupUri() {
+        if let uri = core.getConfiguration("metrics")?.config?.uri {
```
pb82: According to https://github.com/aerogear/proposals/blob/master/sdks/Mobile-configuration.md we should use `url` instead of `uri`.

modules/metrics/metrics/SdkVersion.swift
```diff
@@ -0,0 +1,77 @@
...
+            metricsUri = uri
+        } else {
+             AgsCore.logger.error("Mobile configuration is missing metrics url")
+        }
+    }
+
+    fileprivate func initId() {
+        let defaults = UserDefaults.standard
+        if let versionId = defaults.string(forKey: "metrics-sdk-installation-id")  {
```
pb82: Maybe use a constant for `metrics-sdk-installation-id` ?

modules/metrics/metrics/SdkVersion.swift
```diff
@@ -0,0 +1,77 @@
...
+            clientId=versionId
+            AgsCore.logger.error("clientId! \(clientId)")
+        } else {
+            let uuid = UUID().uuidString
+            clientId = uuid
```
pb82: `clientId` vs. `installationId`, i think we should have consistent naming (will also use installationId on Android).

modules/metrics/metrics/SdkVersion.swift
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    func trackVersion() {
+        if let uri = metricsUri {
+            let payload = SdkVersion(clientId: clientId, sdkVersion: SDK_VERSION, appId: bundleId)
+            AgsCore.logger.debug("Sending metrics \(payload)")
+            let requestDictionary = try? payload.adaptToDictionary()
+            core.getHttp().post(uri, body: requestDictionary, { (response, error) -> Void in
+                if let error = error {
+                    AgsCore.logger.error("An error has occurred when sending app metrics! \(error)")
```
pb82: Just my preference but i would remove the exclamation mark from the message.

modules/metrics/metrics/SdkVersion.swift
wtrocki: We using old format for Android and IOS. New format has url on the top.
Once cmd line tool will be changed and released after sprint, we will adjust formats. 
I'm going to create issue to get that aligned.

modules/metrics/metrics/SdkVersion.swift
wtrocki: +100 - Going to refactor names here as they are not consistent.

modules/metrics/metrics/SdkVersion.swift
wtrocki: Actually going to fix that here: https://issues.jboss.org/browse/AGIOS-648

modules/metrics/AgsMetrics.swift
```diff
@@ -0,0 +1,34 @@
...
+        metrics.append(SdkVersionMetrics(core.getHttp(), config, appData))
+    }
+
+    /**
+     * Force collect metrics for SDK version
+     */
+    open func collectMetrics() {
+        if config.metricsEnabled() {
+            for metricsEngine: Collectable in metrics {
+                // TODO: Use single URL payload vs metrics sending data to individual URLS.
```
wtrocki: @aliok @pb82  - What will be your take on this? Should we have single UDP/HTTP metrics endpoint or each metrics should be as separate endpoint?

From mobile point of view it may be better to just have single endpoint and collect metrics into single payload that is being sent when device is online. WDYT?

wtrocki:
## Retrieving Version in IOS

After experimenting with multiple options like custom cocoapods plugins, plist appenders I decided to include version in core. Version will be changed by release script so there will be no additional overhead for developers to update that.

modules/metrics/metrics/SdkVersion.swift
```diff
@@ -0,0 +1,60 @@
...
+
+    init(_ http: AgsHttpRequest, _ config: MetricsConfig, _ appData: AppData) {
+        self.http = http
+        self.config = config
+        self.appData = appData
+        sdkVersion = AgsCore.getMetadata().SDK_VERSION
+    }
+
+    func collect() {
+        // TODO: Store version in User Preferences and detect that was changed?
```
wtrocki: @aliok @pb82 WDYT? Technically this metrics will change only when user will update application so sending this metrics every time we launch app seems to be battery drainer, antipattern.

SDK version can be saved in preferences and we can detect when it was changed. 
Additionally both IOS and Android have way to detect if application was installed.
However if we agree to have one single endpoint for all "OnStart" metrics then this optimization will not make sense as we could always send this information as metrics payload.

wtrocki:
@danielpassos Apart from couple unanswered questions I think we are in good shape to review this.
Build is failing intentionally - I left TODO statements in code.

modules/metrics/AgsMetrics.swift
aliok: +1 on this. not just when offline though. even when the device is online, we should send stuff in batches into a single URL.
but I wouldn't worry about this right now. perhaps we can start a discussion with the wider group.

modules/metrics/metrics/SdkVersion.swift
aliok: I think sending SDK init events every time is very useful.  Anybody could make use of this data.
Sending once would answer "how many clients use SDK 1.2.3?".
Sending every time would also answer "how many init events happened for SDK 1.2.3 in last N days?" which is basically the how actively an SDK version is used. This is IMO important because there can be 1 installation of SDK 1.0.0 with a very active user, but there can be 100 installations of SDK 0.0.1 with nobody actually using the app.

Also, probably sending one request when app starts won't drain too much battery IMO :)

If we do this, from backend point of view it doesn't change too much. We would just need to adjust the Grafana dashboard so that it doesn't show sdk init event time series data.

modules/metrics/AgsMetrics.swift
aliok: @wtrocki maybe it is better to talk about this once we have more metrics (custom or system). currently all we have is 2 metrics (sdk version and app version) that can be joined in a single payload. but we shouldn't forget this subject.

wtrocki:
@aliok Thanks for responding. Fully agree with you. To summarize:
We are going to send metrics everytime application is launched (initially) and use single payload url that will be provided in `mobile-config.json`

Going to push that to list as well.

modules/metrics/metrics/SdkVersion.swift
pb82: @wtrocki I also agree with @aliok here. Sending the data every time the app is launched is far more useful from a metrics perspective.

modules/metrics/processor/MetricsPublisher.swift
```diff
@@ -0,0 +1,27 @@
...
+     * Allows to publish metrics to external source
+     */
+    func publish(_ payload: MetricsData)
+}
+
+/**
+ * Metrics publisher that logs metrics to console
+ * Can be used when remote server is not available/supported
+ */
+public class MetricsLoggerPublisher: MetricsPublisher {
```
danielpassos: Like I said before I think we should not have more than one class/protocol/whatever in the same file, It means things easy to be found.

modules/metrics/data/MetricsConfig.swift
```diff
@@ -0,0 +1,33 @@
+import AGSCore
+import Foundation
+
+struct MetricsConfigData {
```
danielpassos: Where are we using it?

modules/metrics/data/MetricsConfig.swift
```diff
@@ -0,0 +1,33 @@
...
+}
+
+/**
+ * Wrapper used to manage metrics service configuration
+ */
+class MetricsConfig {
```
danielpassos: I think a separated config class is no need for now since all information is coming from the `ServiceConfig`

modules/metrics/data/AppData.swift
```diff
@@ -0,0 +1,44 @@
+import AGSCore
+import Foundation
+
+/**
+ * Holds static application data that can be used for metrics
+ */
+public class AppData {
```
danielpassos: Just wondering if this kind of class should not be on `core` to be used in other modules or in the app.

modules/metrics/processor/Collectable.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+
+/**
+ * Protocol used for for classes that will colect metrics
+ */
+public protocol Collectable {
```
danielpassos: Why this protocol are in `processor` folder and the implementations are in the `metrics` folder?

danielpassos: I think it should have `metrics` in the name to be more explicit it's a Collectable for Metrics. Makes sense?

modules/metrics/AgsMetrics.swift
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * AeroGear Services metrics
+ */
+open class AgsMetrics: MetricsContainer {
+
+    private let core: AgsCore
+    private let appData: AppData
+    private let config: MetricsConfig
+    private var publisher: MetricsPublisher!
```
danielpassos: What about to have more than one `MetricsPublisher`. I mean what if I wanna have Log and Network or whatever new I write?

modules/metrics/MetricsContainer.swift
```diff
@@ -0,0 +1,16 @@
+import Foundation
+
+/**
+ * Protocol used for mobile metrics management
+ * Allows other SDK and implementations to manage (add) metrics
+ */
+public protocol MetricsContainer {
```
danielpassos: Do we really need a protocol for this? Why not only a simple method in `AgsMetrics`?

example/AeroGearSdkExample/AppDelegate.swift
```diff
@@ -27,6 +29,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     }
 
     func applicationDidBecomeActive(_: UIApplication) {
+        metrics.collectMetrics()
```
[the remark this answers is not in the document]
wtrocki: Outdated. It have it now.

modules/metrics/AgsMetrics.swift
wtrocki: Idea now is to use single one - strategy pattern. If there will be use case for two we could change that and iterate on publishers without hassle. Currently we support only app init event. Once more triggers will appear this functionality may change.

modules/metrics/MetricsContainer.swift
wtrocki: This was created in order to use it in different SDK in order to register metrics.
I'm not sure if this Idea will last, but I wanted to propose some solution for extending metrics without implementing them directly in the metrics SDK.

modules/metrics/data/AppData.swift
wtrocki: +1

modules/metrics/data/MetricsConfig.swift
wtrocki: It's wrapper to get url and future. I wanted to showcase common swift pattern here.
While this class has little code it's good showcase (metrics is first and it will be good to show some good patterns here like optional unwrapping. Initially I had much more code here but decided to simplify that.

modules/metrics/data/MetricsConfig.swift
wtrocki: We getting JSON from mobile-cli and exposing every JSON file to users basing on contract from `mobile-cli` command line tool. 
I get that you do not like to have code that is not used but in case of JSON files there should be exception as this is just part of the API.

modules/metrics/processor/Collectable.swift
wtrocki: metrics folder contains all metrics. "processor' is maybe bad name - I will rename that to `engine`

modules/metrics/processor/MetricsPublisher.swift
wtrocki: +100, Sorry I missed that one

modules/metrics/MetricsContainer.swift
wtrocki: Alternative path will be to just get specific metrics injected on user app.
I think we could talk about couple options here with @pb82 as we will still continue work on this by adding security etc.

danielpassos:
Approval Intense Clapping

aliok:
@wtrocki I know this is merged, but do you mind updating the PR description? Seems outdated and could be confusing in the future.

PR #15, opened 2018-02-09
Adjust metrics to new requirements from proposal

wtrocki:
### Progress

- [x] Add new fields
- [x] Move SDK to singleton
- [x] Expose SDK documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,15 @@ AeroGear Services IOS Swift SDK (AGS)
 1. [End User Getting Started Guide](./docs/getting-started.adoc)
 1. [Service Contributor Guide](./docs/service-guide.adoc)
 
+### List of SDKs
+
+AeroGear Services consist of set of separate SDKs
+
+- [Core](./docs/core/README.adoc)
+Common base for all SDKs
+- [Metrics](./docs/metrics/README.adoc)
+Mobile application metrics SDK
+
 ## Contributing
 
 See [General Contributing Guide](./CONTRIBUTING.md)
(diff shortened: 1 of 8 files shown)


pb82:
I see that you make it possible for users to provide their custom metrics (by implementing MetricsCollectable). I wonder what we can do about default metrics of other SDK modules, say auth. The user could manually add them before sending the metrics. Or maybe there is a way to add them automatically once the module is used?

wtrocki:
@pb82 This interface is designed to be used by other SDK's to provide metrics, but this may imply some dependencies. I think we may just provide NSDictionary based  aproach to avoid strict dependency.  I'm in process of creating tickets to kickstart conversations about it. With that in mind MetricsCollectable can become internal interface at some point, but I would like to run that idea with entire team and community first to get more context.

pb82:
@wtrocki Yes that's the concern: if we want to allow other SDK modules to automatically add their metrics, that requires the interfaces to be part of core and also initializing the SDKs in the right order. Not part of this change however. I  might add a similar abstraction to the Android metrics SDK, but unless we allow user defined metrics there's not really too much use for it.

wtrocki:
@pb82 If you take look into master this prototype is used mostly internally in metrics. This prototype will be also used as adaptable for Dictionary payload. I did not want to go into details for this particular PR as there are couple ways we could do it. I think it's best to have some show and tell to get opinions about it considering that options may require some extensive swift knowledge.

pb82:
@wtrocki Yes agree, we don't have to find a solution for that now.

david-martin:
@wtrocki Apologies for delay in review. I probably can't add much, but looking through the code, it looks reasonable.

wtrocki:
@david-martin No problem. Content of this PR got outdated anyway considering out recent changes on server side proposal, but it should be trivial to adjust anyway.

PR #18, opened 2018-02-22
Metrics refactor

pb82:
### Motivation

WIP: adjust the iOS metrics SDK to the updated proposal and also streamline it's API so that it's closer to the Android SDKs.

Metrics is now part of AGSCore. To use metrics you can either use:

```swift
AgsCore.instance.sendDefaultMetrics()
```

or:

```swift
let metrics: AgsMetrics = AgsCore.instance.getMetricsService()
metrics.publish(MyMetrics())
```


example/AeroGearSdkExample/AppDelegate.swift
```diff
@@ -28,7 +28,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     }
 
     func applicationDidBecomeActive(_: UIApplication) {
-        AgsMetrics.instance.collectMetrics()
+        AgsMetrics.instance.sendDefaultMetrics()
```
wtrocki: If we move to core we could use core object to operate on those to be more aligned with Android

modules/metrics/AgsMetrics.swift
```diff
@@ -48,30 +48,44 @@ open class AgsMetrics: MetricsContainer {
...
     /**
      * Collect metrics for all active metrics collectors
      * Send data using metrics publisher
      */
-    open func collectMetrics() {
-        var metricsPayload: MetricsData = MetricsData()
-        for metricsEngine: MetricsCollectable in metricsCollectors {
-            let engineResult = metricsEngine.collect()
-            metricsPayload.merge(engineResult) { _, new in new }
+    open func sendDefaultMetrics() {
```
wtrocki: That my be misleading. Taking into consideration that there will be change required in android to allow to control default metrics we should avoid that here as well.

modules/metrics/metrics/DeviceMetrics.swift
```diff
@@ -0,0 +1,14 @@
...
+import Foundation
+
+public class DeviceMetrics: MetricsCollectable {
+    public private(set) var identifier: String = "device"
+
+    public func collect() -> MetricsData {
+        return [
+            "platform": "ios",
+            "platformVersion": UIDevice.current.systemVersion,
+            "device": UIDevice.current.model
```
wtrocki: Note to myself. Make sure that android will send device.

modules/metrics/engine/MetricsCollectable.swift
```diff
@@ -4,6 +4,12 @@ import Foundation
  * Protocol used for for classes that will colect metrics
  */
 public protocol MetricsCollectable {
+    /**
+     * A identifier that is used to namespace the metrics data
+     *
+     * @return identifier string
+     */
+    var identifier: String { get }
```
wtrocki: Is that how we do this on android?
I know it's commented, but I'm not sure what that comes from

pb82: Yes, the equivalent on Android is https://github.com/aerogear/aerogear-android-sdk/blob/master/core/src/main/java/org/aerogear/mobile/core/metrics/Metrics.java#L7

The only difference is that on Android it's a method because interfaces can't have properties.

modules/metrics/AgsMetrics.swift
pb82: Do you suggest that we should rename the function?

wtrocki: Yeah. Same suggested on Android PR. Rename it to actual metrics we send.

example/AeroGearSdkExample/AppDelegate.swift
```diff
@@ -28,7 +28,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     }
 
     func applicationDidBecomeActive(_: UIApplication) {
-        AgsMetrics.instance.collectMetrics()
+        AgsCore.instance.sendDefaultMetrics()
```
wtrocki: [Optional] `sentDeviceMetrics` sounds better.

example/AeroGearSdkExample/mobile-services.json
```diff
@@ -37,7 +37,7 @@
       "id": "metrics",
       "name": "metrics",
       "type": "metrics",
-      "url": "https://www.mocky.io/v2/5a79c6572e000028009a5c4f",
+      "url": "http://www.mocky.io/v2/5a79c6572e000028009a5c4f",
```
wtrocki: [Question] Why this change?

modules/metrics/AgsMetrics.swift
wtrocki: Yeah. Suggestions above.

modules/core/metrics/AgsMetrics.swift
```diff
@@ -48,30 +45,44 @@ open class AgsMetrics: MetricsContainer {
...
     }
 
     /**
      * Add new collector to metrics. Collectors allow to collect and append mobile metrics
      * @see Collectable
      * @param collector - new metrics implementation to be added
      */
-    open func addMetricsCollector(_ collector: MetricsCollectable) {
+    private func addMetricsCollector(_ collector: MetricsCollectable) {
         metricsCollectors.append(collector)
```
wtrocki: Entire metrics collectors were designed to work with pull model. When you add some collectors and execute them at some point of time or periodically. Having that now with push model seems to be strange. We could just do publish operation in `enableDefaultMetrics` method. Note this will clash with my suggestion to have them independent above. 
Leaving up to you what is the most preferable.

example/AeroGearSdkExample/mobile-services.json
pb82: Mistake, i used the URL for my local server there but changed it back before pushing the changes. Forgot to change back to https, will fix.

modules/core/metrics/agsmetrics.podspec
```diff
@@ -1,5 +1,6 @@
 Pod::Spec.new do |s|
```
wtrocki: I might miss the motivation here but IMHO that should be removed. Only core will be shippable artifact.

pb82: ah yes, that should be removed. Will update.

modules/core/metrics/AgsMetrics.swift
pb82: Yes agree, that's not really required anymore, we could just do

```swift
func sendDefaultMetrics() {
  publish(AppMetrics())
  publish(DeviceMetrics())
}
```

modules/core/metrics/metrics/AppMetrics.swift
```diff
@@ -0,0 +1,19 @@
+import Foundation
+
+public class AppMetrics: MetricsCollectable {
```
wtrocki: Missing documentation

modules/core/metrics/metrics/DeviceMetrics.swift
```diff
@@ -0,0 +1,13 @@
+import Foundation
+
+public class DeviceMetrics: MetricsCollectable {
```
wtrocki: Missing documentation

modules/core/AgsCore.swift
```diff
@@ -34,6 +36,23 @@ public class AgsCore {
         return http.getHttp()
     }
 
+    public func getMetricsService() -> AgsMetrics {
+        if let metricsInstance = metrics {
+            return metricsInstance
+        } else {
+            metrics = AgsMetrics.init(self)
```
wtrocki: that's really strange way of creating new object. Is that on purpose?
`new AgsMetrics(self)`?

wtrocki:
@pb82 Really sorry for highjacking your PR. I created new PR on upstream to better collaborate without pushing to your fork. All my comments were addressed there already to save you troubles.

PR #19, opened 2018-02-22
 📝  make sure `pod install` is executed first before opening the xcode

wei-lee:
project

### Motivation

By default the `AeroGearSdkExample.xcworkspace` doesn't exist In the repo.  Developers need to run `pod install` first

### Description

Update the instructions to make it clear that `pod install` needs to be executed first

### Progress

- [x] Update instruction

### DIFF
--- a/docs/service-guide.adoc
+++ b/docs/service-guide.adoc
@@ -98,7 +98,9 @@ SDK and example applications can be developed in XCode.
 To open xcode project Execute
 
 ----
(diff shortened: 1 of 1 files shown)


wtrocki:
@wei-lee Another good hint is to execute pod install when you have changed something in the top level api. 
For example when renaming method XCode may not see that immediately. 
Pod install saves 100% of the problems.

PR #20, opened 2018-02-22
Metrics refactor

wtrocki:
### Motivation

Adjust the iOS metrics SDK to the updated proposal and also streamline it's API so that it's closer to the Android SDKs.

Metrics is now part of AGSCore. To use metrics you can either use:

```swift
let metrics = AgsCore.instance.getMetrics()
metrics.publish(MyMetrics())
```
## Notes

@pb82 This is update for all my comments from your PR. I could not manage to push changes to your fork.
Created upstream branch for better collaboration.

See https://github.com/aerogear/aerogear-ios-sdk/pull/18


wtrocki:
@pb82 `./scripts/develop.sh` script can be used locally to format and fix style issues.

docs/metrics/README.adoc
```diff
@@ -19,7 +19,7 @@ Add to your app delegate following code
 [source,swift]
 ----
     func applicationDidBecomeActive(_: UIApplication) {
-        AgsMetrics.instance.collectMetrics()
+        AgsCore.instance.sendAppDeviceMetrics()
```
danielpassos: Can we go with `sendAppAndDeviceMetrics` only to use the same I have suggested on Android?

modules/core/AgsCore.swift
```diff
@@ -34,6 +37,24 @@ public class AgsCore {
...
+     */
+    public func getMetrics() -> MetricsContainer {
+        return metrics
+    }
+
+    /**
+     * Send default app and device metrics including a unique device identifier and
+     * App and SDK version information
+     */
+    public func sendAppDeviceMetrics() {
```
danielpassos: I know it's a convenient delegate for send "default" metrics but I personally don't like to have it. It's not about correct on wrong it's only a personal opinion I think all metrics stuff should be raised directly no `AgsMetrics`

modules/core/metrics/AgsMetrics.swift
```diff
@@ -0,0 +1,72 @@
...
+     */
+    public func setMetricsPublisher(_ publisher: MetricsPublisher) {
+        self.publisher = publisher
+    }
+
+    /**
+     * Collect metrics for all active metrics collectors
+     * Send data using metrics publisher
+     */
+    open func sendAppDeviceMetrics() {
```
danielpassos: Can we go with `sendAppAndDeviceMetrics` only to use the same I have suggested on Android?

modules/core/metrics/MetricsContainer.swift
```diff
@@ -0,0 +1,27 @@
+import Foundation
+
+/**
+ * Protocol used for mobile metrics management
+ * Allows other SDK and implementations to manage (add) metrics
+ */
+public protocol MetricsContainer {
```
danielpassos: @wtrocki @pb82 Do we still need to have `MetricsContainer`? Why `AgsMetrics` can't handler directly with `MetricsPublisher`?

modules/core/metrics/engine/MetricsCollectable.swift
```diff
@@ -4,6 +4,12 @@ import Foundation
  * Protocol used for for classes that will colect metrics
  */
 public protocol MetricsCollectable {
```
danielpassos: Should we rename it to only `Metrics` to align with the Android one, or rename the Android one for `MetricsCollectable`?

modules/core/metrics/MetricsContainer.swift
wtrocki: Metrics container is just protocol that defines our front operations. 
This will mean that if we want to have public methods that will not be exposed we will be able to do it.
Protocol is also safeguard as when you change protocol you are actually breaking api for the users.

david-martin:
@wtrocki I'd like to try these changes out, but that will probably be tomorrow some time (when I have a mac available)

modules/core/metrics/MetricsContainer.swift
danielpassos: I see what it's doing but I'm still not 100% sure about that. Anyway I will allow it

wtrocki: @pb82 What is your opinion on this. Maybe I'm trying to do something that makes no sense.
I was following some examples to write frameworks

pb82: I think that the `MetricsContainer` not needed, at least for now marking methods as public or private should give us enough isolation. Even if we wanted the Protocol, the name is a bit confusing: `MetricsContainer` sounds like it's a container type when it's actually an interface. Something like `MetricsService` would be better.

ptrocki: ` at least for now marking methods as public or private should give us enough isolation` 
- Yes but protocol is better for isolation and API purposes. 
`Even if we wanted the Protocol, the name is a bit confusing:`
- True, the postfix should be `-able` for protocols.

modules/core/metrics/AgsMetrics.swift
```diff
@@ -0,0 +1,72 @@
...
+            payload[metric.identifier] = result
+        }
+        payload = metricsRoot().merging(payload) { orig, _ in orig }
+        publisher.publish(payload)
+    }
+
+    private func metricsRoot() -> [String: Any] {
+        return [
+            "clientId": appData.clientId,
+            "timestamp": "\(NSDate().timeIntervalSince1970 * 1000)",
```
david-martin: @wtrocki I think this should be a number, without an decimal places.
Locally, I'm seeing a value like this
```
"timestamp": "15169379639166.49"
```

@darahayes What does the server validate this field as?
fyi, it's at the top of the metrics object hierarchy, along with `clientId`.

@wtrocki we should standardise across all clients on the structure and expected field types, according to what the server expects

darahayes: @david-martin. Right now the client timestamp is not being saved in the database. It is being parsed by the server into an integer though. In postgres the timestamps are being stored as a timestampz type.

david-martin: @darahayes will the server reject a request where the timestamp is a string, or has decimal precision?
I think it should, and only a number should be accepted for that field

darahayes: At the moment there is basically no validation of the actual values coming from the client. I focused on getting the data into the database first. I think maybe we could open another ticket around server side validation.

PR #21, opened 2018-02-23
AGIOS-653 - documentation for metrics

wtrocki:
### Motivation
 
Documentation updates for metrics

### DIFF
--- a/docs/metrics/README.adoc
+++ b/docs/metrics/README.adoc
@@ -8,24 +8,22 @@ AeroGear Services metrics
 1. SDK and application version 
 Tracking AeroGear Service SDK version and Application version
 +
-1. Application started timestamp metrics
-Metrics sending current timestamp 
+1. Device metrics
+Metrics sending device and platform information
 
+== Sending app and device metrics
 
-== Application usage
-
-Add to your app delegate following code
-
+To send bundled application metrics add to following code to main application delegate:
 [source,swift]
 ----
-    func applicationDidBecomeActive(_: UIApplication) {
-        AgsCore.instance.sendAppDeviceMetrics()
-    }
+func applicationDidBecomeActive(_: UIApplication) {
+    AgsCore.instance.sendAppAndDeviceMetrics()
+}
 ----
 
-== Contributing to metrics SDK 
+== Adding new metrics 
 
-1.Create new metrics collector by extending `MetricsCollectable` prototype
+1.Create new metrics collector by extending `Metrics` prototype
 [source,swift]
 ----
  public class MyMetrics: Metrics {
@@ -37,5 +35,17 @@ Add to your app delegate following code
 2. Append metrics to metrics container  
 [source,swift]
 ----
-    AgsCore.instance.getMetrics().publish(MyMetrics())
+AgsCore.instance.getMetrics().publish(MyMetrics())
 ----
+
+ == Replacing default metrics publisher
+
+Developers can provide custom publishers to change the way metrics are sent or stored:
+
+[source,swift]
+----
+MyPublisher publisher: MetricsPublisher = MyPublisher()
+AgsCore.instance.getMetrics().setMetricsPublisher(publisher) 
+----
+
+
--- a/docs/metrics/README.adoc
+++ b/docs/metrics/README.adoc
@@ -1,51 +1,24 @@
-= AeroGear Services metrics documentation
+= AeroGear Services Metrics documentation
 
-AeroGear Services metrics
+This document explains how to use the AeroGear Metrics SDK on the IOS platform and integrate it with an App.
 
+== Provided metrics
 
-== Internal metrics
+The default metrics contain App metrics, Device metrics and possibly other types in the future.
+A Client ID and a timestamp are added to the top level of the Metrics object for categorization and visualization purposes. 
+The Client ID will be generated and stored on the device when the App is started the first time.
 
-1. SDK and application version 
-Tracking AeroGear Service SDK version and Application version
-+
-1. Device metrics
-Metrics sending device and platform information
+=== App metrics
 
-== Sending app and device metrics
+- *App ID*: The package name of your App
+- *App version*: The version of your App
+- *SDK version*: The version of the Aerogear SDK that is used
 
-To send bundled application metrics add to following code to main application delegate:
-[source,swift]
-----
-func applicationDidBecomeActive(_: UIApplication) {
-    AgsCore.instance.sendAppAndDeviceMetrics()
-}
-----
+=== Device metrics
 
-== Adding new metrics 
+- *Platform*: Hardcoded to `ios`
+- *Platform Version*: Version of the platform (IOS10 etc.)
 
-1.Create new metrics collector by extending `Metrics` prototype
-[source,swift]
-----
- public class MyMetrics: Metrics {
-    public func collect() -> MetricsData {
-        return ["mylabel","my value"]
-    }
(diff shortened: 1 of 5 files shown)


docs/metrics/README.adoc
```diff
@@ -8,24 +8,22 @@ AeroGear Services metrics
 1. SDK and application version 
 Tracking AeroGear Service SDK version and Application version
 +
-1. Application started timestamp metrics
-Metrics sending current timestamp 
+1. Device metrics
+Metrics sending device and platform information
```
danielpassos: device and platform or device and app?

docs/metrics/README.adoc
```diff
@@ -8,24 +8,22 @@ AeroGear Services metrics
...
+func applicationDidBecomeActive(_: UIApplication) {
+    AgsCore.instance.sendAppAndDeviceMetrics()
+}
 ----
 
-== Contributing to metrics SDK 
+== Adding new metrics 
 
-1.Create new metrics collector by extending `MetricsCollectable` prototype
+1.Create new metrics collector by extending `Metrics` prototype
```
danielpassos: s/Create new metrics collector/Create new metrics

docs/metrics/README.adoc
```diff
@@ -37,5 +35,17 @@ Add to your app delegate following code
 2. Append metrics to metrics container  
 [source,swift]
 ----
-    AgsCore.instance.getMetrics().publish(MyMetrics())
-----
\ No newline at end of file
+AgsCore.instance.getMetrics().publish(MyMetrics())
+----
+
+ == Replacing default metrics publisher
```
danielpassos: I really like to have this on doc!!! :sparkling_heart:

wtrocki:
@danielpassos Could you merge that for me if you are ok with the changes?

docs/metrics/README.adoc
wtrocki: This has only device and platform. Nothing else.

wtrocki:
Happy to apply any suggestions. Get your point on the splitting to different personas - do not have any strong opinion about this. 
I followed what we have on Android: https://github.com/aerogear/aerogear-android-sdk/blob/master/docs/getting-started/metrics.adoc

Decided to skip `Using metrics` chapter as this is common for any SDK, but if you want I could copy it here for better alignment. 

I see no problem with splitting some elements to contributor guide and user guide, but at this point I do not have any opinions - publishers are typically for end users to modify metrics to their internal company security requirements. Metrics can be used both by SDK contributors and end users. 

@david-martin 
What I mean is just that a there is no straight line here between contribution and user story, until we really want to have one.

david-martin:
@wtrocki Lets draw that hard line now so.

If the person is just consuming the SDK, they will need 'end user' docs.
If the person is contributing to the SDK, they will need 'contributing' docs.

Contributing docs are unlikely to end up in 'customer' or 'productised' docs.
Lets split them please.
@finp Is this in line with your thinking? I'm not sure if we will have that split on aerogear.org too?

wtrocki:
@pb82 @finp Any voices against having that separate? I would like to do it for the both platforms so we will have them aligned.

pb82:
@wtrocki I'm ok with the split but we should implement this format for all SDKs. Looks like the JS SDK already has this docs separation.

wtrocki:
Great! Doing it!

docs/metrics/README.adoc
```diff
@@ -1,41 +1,38 @@
-= AeroGear Services metrics documentation
+= AeroGear Services Metrics documentation
 
-AeroGear Services metrics
+This document explains how to use the AeroGear Metrics SDK on the Android platform and integrate it with an App.
```
danielpassos: Android?! Copy and Past? :disappointed:

docs/metrics/README.adoc
```diff
@@ -1,41 +1,38 @@
...
+All you have to do is to make sure that your `mobile-services.json` contains a configuration section for it.
+You can verify this with the mobile cli:
 
-1. SDK and application version 
-Tracking AeroGear Service SDK version and Application version
-+
-1. Application started timestamp metrics
-Metrics sending current timestamp 
+```
+$ mobile get clientconfig <app id> --namespace=<project> -o json
```
darahayes: Is this how an end user will officially use the mobile CLI? or is this some shortcut thing we're doing at the moment?

darahayes:
@wtrocki I left one comment around the mobile cli usage.

I'm not going to leave an official approval because I'm not an expert in this area but I think it looks pretty good. Unofficial +1 from me.

docs/metrics/README.adoc
danielpassos: I think we should omit how to get the JSON file or in the future link this to the cli doc

wtrocki: This was added on Android by docs team. Was just following the Android path.

docs/metrics/README.adoc
```diff
@@ -1,41 +1,38 @@
...
+```
 
+The result should include a `metrics` section inside the `services` array.
+If no such configuration is provided then the Metrics module will not send any data over the network but will still log the data to the console and you can view them in Logcat or another remote logging solution.
```
pb82: Logcat is Android specific, we should remove that for the iOS docs.

docs/metrics/README.adoc
```diff
@@ -1,41 +1,38 @@
...
-+
-1. Application started timestamp metrics
-Metrics sending current timestamp 
+```
+$ mobile get clientconfig <app id> --namespace=<project> -o json
+```
 
+The result should include a `metrics` section inside the `services` array.
+If no such configuration is provided then the Metrics module will not send any data over the network but will still log the data to the console and you can view them in Logcat or another remote logging solution. 
+Once you initialize the core SDK some default metrics will be collected and sent.
```
pb82: That is no longer the case right? `sendAppAndDeviceMetrics` needs to be called explicitely by the user. We need to change that in the Android docs too.

docs/metrics/extending-metrics.adoc
```diff
@@ -0,0 +1,44 @@
...
+
+== Adding new metrics 
+
+Users can also create their own metrics by extending `Metrics` prototype. 
+
+1.Create new metrics by extending `Metrics` prototype
+[source,swift]
+----
+ public class MyMetrics: Metrics {
+    public func collect() -> MetricsData {
```
pb82: The `identifier` property is missing

wtrocki:
@pb82 Aligned that with Android documentation and in results got some outdated content. Fixed all of that now. Awesome reviews guys!

docs/metrics/README.adoc
```diff
@@ -1,41 +1,24 @@
...
 
-1. SDK and application version 
-Tracking AeroGear Service SDK version and Application version
-+
-1. Application started timestamp metrics
-Metrics sending current timestamp 
+=== App metrics
 
+- *App ID*: The package name of your App
+- *App version*: The version of your App as defined in the top level `build.gradle`
```
pb82: `build.gradle` is Android specific

david-martin:
@wtrocki @pb82 @danielpassos Good to merge this now?
Looks like a rebase is needed.

I'll be following up with @finp on any changes required based on ongoing discussion in https://github.com/aerogearcatalog/metrics-apb/pull/29

pb82:
@david-martin i'm happy to get this merged. Looks good now, all comments adressed.

PR #22, opened 2018-02-23
Small fix to docs

josemigallas:
### Description
Services names in getting-started guide are outdated.

### DIFF
--- a/docs/getting-started.adoc
+++ b/docs/getting-started.adoc
@@ -51,14 +51,14 @@ pod install
 
 == Getting Started with Swift
 
-1) Import the AgsCore in the application .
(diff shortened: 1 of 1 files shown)


wtrocki:
@josemigallas That is blocked by actual release. We haven't released that yet. 
If you are interested in contribution take look on contributing guide. 
User guide is just template that will need to be adjusted when we release. 
We have ticket to cover this work and then we will provide actual names

danielpassos:
+1 on what @wtrocki we should close that for now and revisit in the future

josemigallas:
@wtrocki @danielpassos We could keep it coherent in the meanwhile

PR #23, opened 2018-02-26
Make configuration immutable

wtrocki:
### Motivation

Change structures to be immutable.


wtrocki:
Fixing build

wei-lee:
we should make the value immutable wherever possible.

wtrocki:
Fixing build. Going to review code for more cases where we use var.

wtrocki:
Fixed build by removing default value. I verified that using sample app.
Thanks for review.

PR #27, opened 2018-02-28
[WIP] Idm sdk base

wei-lee:
### Motivation

https://issues.jboss.org/browse/AEROGEAR-2086

### Description

Create some of the base classes/interfaces for the iOS IDM SDK based on the android one.

### Progress

- [x] Add classes
- [x] Fix code style

Please see commit ca691d9. The other one is auto generated by swiftlint.

ping @aidenkeating @rachael-oregan @ziccardi could you take a look at this PR if you are planning on the ios IDM SDK. We will chat about it during the standup today.

### DIFF
--- a/modules/auth/AgsAuth.swift
+++ b/modules/auth/AgsAuth.swift
@@ -9,28 +9,27 @@ open class AgsAuth {
     let credentialManager: CredentialsManager
     let keycloakConfig: KeycloakConfig
     let authServiceConfig: AuthenticationConfig
-    
+
     init(serviceConfig: ServiceConfig, authConfig: AuthenticationConfig) {
         keycloakConfig = KeycloakConfig(serviceConfig)
         authServiceConfig = authConfig
         credentialManager = CredentialsManager()
         authenticator = OIDCAuthenticator()
     }
-    
+
     public func login(presentingViewController: UIViewController, onCompleted: @escaping (User?, Error?) -> Void) {
-        
+
     }
-    
+
     public func resumeAuth(url: URL) -> Bool {
         return false
     }
-    
+
     public func logout(onCompleted: @escaping (Error?) -> Void) {
-        
+
     }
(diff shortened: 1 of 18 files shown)


modules/auth/AgsAuth.swift
```diff
@@ -0,0 +1,35 @@
...
+        authServiceConfig = authConfig
+        credentialManager = CredentialsManager()
+        authenticator = OIDCAuthenticator()
+    }
+
+    public func login(presentingViewController: UIViewController, onCompleted: @escaping (User?, Error?) -> Void) {
+
+    }
+
+    public func resumeAuth(url: URL) -> Bool {
```
aidenkeating: What is this method for?

wei-lee: similar to the authentication flow on android, when user logged in via the browser, the app will be resumed. This method should be used when the app is resumed to completed the login process.

aidenkeating: Ah yep, makes sense now, thanks

wtrocki:
@wei-lee You will need to install swift lint on your machine. See ./scripts/develop.sh for details.

modules/auth/authenticator/Authenticator.swift
```diff
@@ -0,0 +1,10 @@
+//
+//  Authenticator.swift
+//  AGSAuth
+
+import Foundation
+
+public protocol Authenticator {
```
wtrocki: Good style is to have protocols ending ..Able. Not big deal. Just mentioning

wei-lee: I don't know if that's possible for all the protocols?

wtrocki: Yeah. It will be something like `Authenticable` and that sounds strange! Let's ignore this.

PR #28, opened 2018-03-01
 🎉 implement the logout function

wei-lee:
### Motivation

https://issues.jboss.org/browse/AEROGEAR-2088

### Description

Add implementation for the logout function in OIDCAuthenticator

### Progress

- [x] Implementation
- [x] Unit tests

ping @ziccardi @rachael-oregan @aidenkeating

### DIFF
--- a/example/AeroGearSdkExampleTests/mocks/MockHttpRequest.swift
+++ b/example/AeroGearSdkExampleTests/mocks/MockHttpRequest.swift
@@ -1,0 +1,58 @@
+//
+//  MockHttpRequest.swift
+//  AeroGearSdkExampleTests
+//
+//  Created by Wei Li on 01/03/2018.
+//  Copyright © 2018 AeroGear. All rights reserved.
+//
+
+import Foundation
+import AGSCore
+
+enum  MockHttpErrors : Error {
+    case NetworkError
+}
+
+class MockHttpRequest: AgsHttpRequestProtocol {
+    var dataForGet: Any?
+    var errorForGet: Error?
+    var dataForPost: Any?
+    var errorForPost: Error?
+    var dataForPut: Any?
+    var errorForPut: Error?
+    var dataForDelete: Any?
+    var errorForDelete: Error?
+    
+    func get(_ url: String, params: [String : AnyObject]?, headers: [String : String]?, _ handler: @escaping (Any?, Error?) -> Void) {
+        if let err = errorForGet {
+            handler(nil, err)
+        } else {
+            handler(dataForGet, nil)
+        }
+    }
+    
+    func post(_ url: String, body: [String : Any]?, headers: [String : String]?, _ handler: @escaping (Any?, Error?) -> Void) {
+        if let err = errorForPost {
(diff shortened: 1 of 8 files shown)


modules/auth/config/KeycloakConfig.swift
```diff
@@ -22,4 +22,9 @@ class KeycloakConfig {
             """)
         }
     }
+
+    func getLogoutUrl(idToken: String) -> String {
```
wei-lee: @rachael-oregan FYI,  new method

modules/auth/credentials/CredentialsManager.swift
```diff
@@ -4,13 +4,24 @@
...
     func load() -> OIDCCredentials? {
         return nil
     }
 
     func save(credentials: OIDCCredentials) {
 
     }
+
+    /** Remove local credentials*/
+    func clear() {
```
wei-lee: @aidenkeating FYI, new method

wei-lee:
@danielpassos @wtrocki mind taking a look at this PR and approve it if that is ok?

modules/auth/authenticator/OIDCAuthenticator.swift
```diff
@@ -3,13 +3,36 @@
...
+        self.keycloakConfig = keycloakConfig
+        self.authConfig = authConfig
+        self.credentialManager = credentialManager
+    }
+
     public func authenticate(presentingViewController: UIViewController, onCompleted: @escaping (User?, Error?) -> Void) {
         //Add implementation
     }
 
     public func logout(currentUser: User, onCompleted: @escaping (Error?) -> Void) {
```
aidenkeating: Should we be adding documentation to these functions?

wei-lee: yes, we should. Add them now

example/AeroGearSdkExampleTests/mocks/MockHttpRequest.swift
```diff
@@ -0,0 +1,55 @@
...
+//
+
+import Foundation
+import AGSCore
+
+enum  MockHttpErrors : Error {
+    case NetworkError
+}
+
+class MockHttpRequest: AgsHttpRequestProtocol {
```
wtrocki: Maybe we could bring something like https://github.com/Brightify/Cuckoo later. WDYT? Is it worth to add some mocking lib to the app?

wei-lee: There is a bit overhead of setting it up, and I am not sure how much value we can get from it TBH. Maybe later on if we have more classes we need to mock, perhaps we can give it a try. For the moment, I don't think it worth it. But it is just my thought, I am happy to go with either way.

wtrocki: I did not mean to introduce that in this PR - but was just curious if you find the value of using it. Perfectly fine then. No additional action needed.

PR #31, opened 2018-03-02
Implementing login

ziccardi:
### Motivation

Implementing login

JIRA: [https://issues.jboss.org/browse/AGDROID-AEROGEAR-2087]

### Description

Login implementation

### Progress

- [x] Implement login
- [x] Add comments
- [x] Add tests


modules/auth/authenticator/OIDCAuthenticator.swift
```diff
@@ -23,6 +27,36 @@ public class OIDCAuthenticator: Authenticator {
...
+        let oidServiceConfiguration = OIDServiceConfiguration(authorizationEndpoint: self.keycloakConfig.authenticationEndpoint!, tokenEndpoint: self.keycloakConfig.tokenEndPoint!)
+        let oidAuthRequest = OIDAuthorizationRequest(configuration: oidServiceConfiguration, clientId: self.keycloakConfig.clientId!, scopes: [OIDScopeOpenID, OIDScopeProfile], redirectURL: authConfig.redirectURL, responseType: OIDResponseTypeCode, additionalParameters: nil)
+        
+        let appDelegate = UIApplication.shared.delegate as! AppDelegate
+        appDelegate.authService = self;
+        
+        //this will automatically exchange the token to get the user info
+        self.currentAuthorisationFlow = OIDAuthState.authState(byPresenting: oidAuthRequest, presenting: viewController) {
+            authState,error in
+            if authState != nil && authState?.authorizationError == nil {
```
wtrocki: This code is too Java like. 
Avoid `authState != nil` - this can be unwrapped by `guard` of `if`

modules/auth/authenticator/OIDCAuthenticator.swift
```diff
@@ -23,6 +27,36 @@ public class OIDCAuthenticator: Authenticator {
         //Add implementation
     }
 
+    func performAuthentication(presentingViewController viewController:UIViewController, onCompleted: @escaping (User?, Error?) -> Void) {
+        self.onCompleted = onCompleted
+        let oidServiceConfiguration = OIDServiceConfiguration(authorizationEndpoint: self.keycloakConfig.authenticationEndpoint!, tokenEndpoint: self.keycloakConfig.tokenEndPoint!)
```
wtrocki: This code can blow up. We should not use `!` to unwrap optionals.
self.keycloakConfig.authenticationEndpoint!

modules/auth/authenticator/OIDCAuthenticator.swift
```diff
@@ -23,6 +27,36 @@ public class OIDCAuthenticator: Authenticator {
...
+        let oidAuthRequest = OIDAuthorizationRequest(configuration: oidServiceConfiguration, clientId: self.keycloakConfig.clientId!, scopes: [OIDScopeOpenID, OIDScopeProfile], redirectURL: authConfig.redirectURL, responseType: OIDResponseTypeCode, additionalParameters: nil)
```
wtrocki: `additionalParameters: nil` - that seems to be optional

> https://github.com/kamaldheeraj/AppAuthSampleSwift/blob/master/Pods/AppAuth/Source/OIDAuthorizationRequest.h#L178

modules/auth/config/KeycloakConfig.swift
```diff
@@ -9,11 +9,16 @@ import AGSCore
 class KeycloakConfig {
     private let sdkId = "keycloak"
 
+    var clientId: String?
```
wtrocki: Suggestion is to make some of the required fields non optionals and just do proper error handling when they are missing.

modules/auth/authenticator/OIDCAuthenticator.swift
```diff
@@ -23,6 +27,36 @@ public class OIDCAuthenticator: Authenticator {
...
+        
+        let appDelegate = UIApplication.shared.delegate as! AppDelegate
+        appDelegate.authService = self;
```
wtrocki: [Question] Why we do that this way? Could we just have this instance in app delegate (or even do singleton if that makes sense for SDK).

modules/auth/authenticator/OIDCAuthenticator.swift
```diff
@@ -12,17 +13,75 @@ public class OIDCAuthenticator: Authenticator {
...
+                if let err = state.authorizationError {
+                    self.authFailure(authState: state, error: err)
+                } else {
+                    self.authSuccess(authState: state)
+                }
+            } else {
+                self.authFailure(error: error)
+            }
+            
+            if authState != nil && authState?.authorizationError == nil {
```
wei-lee: I think this block is a duplication of the above block, we only need to keep one of them?

modules/auth/authenticator/OIDCAuthenticator.swift
```diff
@@ -47,8 +106,25 @@ public class OIDCAuthenticator: Authenticator {
...
+        }
+    }
+    
     public func resumeAuth(url: URL) -> Bool {
         return false
     }
+
+}
+
+func base64urlToBase64(base64url: String) -> String {
```
wei-lee: I don't think this is required

modules/auth/authenticator/OIDCAuthenticator.swift
```diff
@@ -47,8 +106,25 @@ public class OIDCAuthenticator: Authenticator {
...
     }
-
+    
+    func authCompleted(identity: User?, error: Error?) {
+        if (self.onCompleted != nil) {
+            self.onCompleted!(identity, error)
+        }
+    }
+    
     public func resumeAuth(url: URL) -> Bool {
```
wei-lee: can we add implementation for this method? It should be easy to implement.

modules/auth/authenticator/OIDCAuthenticator.swift
```diff
@@ -12,17 +13,75 @@ public class OIDCAuthenticator: Authenticator {
...
+            } else {
+                self.authFailure(authState: authState, error: error)
+            }
+        }
     }
-
+    
+    fileprivate func assignAuthState(authState: OIDAuthState?) {
+        self.authState = authState
+        self.identity = self.getIdentity(authState: authState)
```
wei-lee: I don't think we really need these properties. We can just use parameters and pass them around.

modules/auth/authenticator/OIDCAuthenticator.swift
```diff
@@ -12,17 +13,75 @@ public class OIDCAuthenticator: Authenticator {
...
+            let credentials = OIDCCredentials(state: state)
+            return User(credential: credentials, clientName: self.keycloakConfig.clientID)
+        }
+        
+        return nil
+    }
+    
+    func authSuccess(authState: OIDAuthState) {
+        self.assignAuthState(authState: authState)
+        self.authCompleted(identity: self.identity, error: nil)
```
wei-lee: I think the implementation here can be improved a little. I don't think we need this many functions:

```
let credentials = OIDCCredentials(state: state)
if let user = User(credential: credentials, clientName: self.keycloakConfig.clientID) {
    credentialManager.save(credentials)
    self.authCompleted(user: user, error: nil)
} else {
    credentialManager.clear()
    let err = someError
    self.authCompleted(user: nil, error: err)
}
```

modules/auth/authenticator/OIDCAuthenticator.swift
```diff
@@ -12,17 +13,75 @@ public class OIDCAuthenticator: Authenticator {
...
+    }
+    
+    func authFailure(authState: OIDAuthState? = nil, error: Error?) {
+        let e = authState?.authorizationError ?? error
+        self.assignAuthState(authState: nil)
+        self.authCompleted(identity: nil, error: e)
```
wei-lee: same here, all we need to do is:

```
credentialManager.clear()
self.authCompleted(user: nil, error: e)
```

modules/auth/authenticator/OIDCAuthenticator.swift
```diff
@@ -47,8 +88,19 @@ public class OIDCAuthenticator: Authenticator {
...
-
+    
+    func authCompleted(identity: User?, error: Error?) {
+        if (self.onCompleted != nil) {
+            self.onCompleted!(identity, error)
+        }
+    }
+    
     public func resumeAuth(url: URL) -> Bool {
+        if self.currentAuthorisationFlow!.resumeAuthorizationFlow(with: url) {
```
wei-lee: this will thrown an exception if `currentAuthorisationFlow` is nil. Better to do it this way:

```
guard let flow = self.currentAuthorisationFlow else {
    return false
}
let result = flow.resumeAuthorizationFlow(with: url)
self. currentAuthorisationFlow = nil
return result
```

modules/auth/authenticator/OIDCAuthenticator.swift
```diff
@@ -12,17 +13,57 @@ public class OIDCAuthenticator: Authenticator {
...
+                    self.authSuccess(authState: state)
+                }
+            } else {
+                self.authFailure(error: error)
+            }
+        }
     }
-
+    
+    fileprivate func getIdentity(authState: OIDAuthState?) -> User? {
```
wei-lee: we can make the parameter not optional.

modules/auth/authenticator/OIDCAuthenticator.swift
```diff
@@ -12,17 +13,57 @@ public class OIDCAuthenticator: Authenticator {
...
+            
+            if let state = authState {
+                if let err = state.authorizationError {
+                    self.authFailure(authState: state, error: err)
+                } else {
+                    self.authSuccess(authState: state)
+                }
+            } else {
+                self.authFailure(error: error)
+            }
```
wei-lee: this works fine, just a little bit too much `if else`. A simple way to improve this is to use `guard`. So this can be written as the following:

```
guard let state = authState else {
    self.authFailure(error: error)
}
if let err = state.authorizationError {
    self.authFailure(authState: state, error: err)
} else {
    self.authSuccess(authState: state)
}
```

modules/auth/authenticator/OIDCAuthenticator.swift
```diff
@@ -12,17 +13,58 @@ public class OIDCAuthenticator: Authenticator {
...
         self.authConfig = authConfig
         self.credentialManager = credentialManager
     }
 
-    public func authenticate(presentingViewController _: UIViewController, onCompleted _: @escaping (User?, Error?) -> Void) {
-        // Add implementation
+    public func authenticate(presentingViewController: UIViewController, onCompleted: @escaping (User?, Error?) -> Void) {
+        //self.onCompleted = onCompleted
+        let oidServiceConfiguration = OIDServiceConfiguration(authorizationEndpoint: self.keycloakConfig.authenticationEndpoint, tokenEndpoint: self.keycloakConfig.tokenEndpoint)
+        let oidAuthRequest = OIDAuthorizationRequest(configuration: oidServiceConfiguration, clientId: self.keycloakConfig.clientID, scopes: [OIDScopeOpenID, OIDScopeProfile], redirectURL: authConfig.redirectURL, responseType: OIDResponseTypeCode, additionalParameters: nil)
```
wtrocki: [Optional] Loads of parameters here. Best to format that by adding line breaks

modules/auth/authenticator/OIDCAuthenticator.swift
```diff
@@ -12,17 +13,58 @@ public class OIDCAuthenticator: Authenticator {
...
+        
+        //this will automatically exchange the token to get the user info
+        self.currentAuthorisationFlow = OIDAuthState.authState(byPresenting: oidAuthRequest, presenting: presentingViewController) {
+            authState,error in
+            
+            if let state = authState {
+                if let err = state.authorizationError {
+                    self.authFailure(error: err, onCompleted: onCompleted)
```
wtrocki: It seems that in auth failure we do not care what error is send so this logic can become

modules/auth/authenticator/OIDCAuthenticator.swift
```diff
@@ -12,17 +13,58 @@ public class OIDCAuthenticator: Authenticator {
...
         self.http = http
         self.keycloakConfig = keycloakConfig
         self.authConfig = authConfig
         self.credentialManager = credentialManager
     }
 
-    public func authenticate(presentingViewController _: UIViewController, onCompleted _: @escaping (User?, Error?) -> Void) {
-        // Add implementation
+    public func authenticate(presentingViewController: UIViewController, onCompleted: @escaping (User?, Error?) -> Void) {
+        //self.onCompleted = onCompleted
```
wtrocki: Let's remove that

modules/auth/authenticator/OIDCAuthenticator.swift
wei-lee: not sure I get what you mean. We do propagate the error to the callback.

wtrocki: I meant that we can try to simplify that code. I have tried that with @aidenkeating example and it it's not the case

wtrocki:
@ziccardi Do you need any help with unit tests?

ziccardi:
@wei-lee Mind giving another look? I added some test for the `login` function, but I could add it only for failed login. Succesful login would need an `AuthState` object, but it is not an interface and the init method is not overridable (at that is what I discovered by trying to mock it).

example/AeroGearSdkExampleTests/auth/authenticator/OIDCAuthenticatorTest.swift
```diff
@@ -6,9 +6,37 @@
...
 
 class OIDCAuthenticatorTest: XCTestCase {
+    
+    class MockOIDCAuthenticator : OIDCAuthenticator {
+        
+        let fail: Bool
+        
+        init(http: AgsHttpRequestProtocol, keycloakConfig: KeycloakConfig, authConfig: AuthenticationConfig, credentialManager: CredentialManagerProtocol, fail: Bool = false) {
+            self.fail = fail
+            super.init(http: http, keycloakConfig: keycloakConfig, authConfig: authConfig, credentialManager: credentialManager)
```
wei-lee: normally `super.init` should be called first?

ziccardi: That's what I thought too... but looks like swift works differently that android and wants the extra variables to be initialised first

example/AeroGearSdkExampleTests/auth/authenticator/OIDCAuthenticatorTest.swift
```diff
@@ -27,7 +55,20 @@ class OIDCAuthenticatorTest: XCTestCase {
...
+        func cancel() {
+        }
+        func resumeAuthorizationFlow(with URL: URL) -> Bool {
+            return false
+        }
+        func failAuthorizationFlowWithError(_ error: Error) {
+        }
+    }
+    
+    class MockUIViewController : UIViewController {
```
wei-lee: can we move the mocks to the `mocks` folder to make the code a little bit cleaner?

modules/auth/authenticator/OIDCAuthenticator.swift
```diff
@@ -2,27 +2,92 @@
...
+                }
+            } else {
+                callback(nil, error)
+            }
+        })
+    }
+    
+    func authSuccess(credentials: OIDCCredentials, onCompleted: @escaping (User?, Error?) -> Void) {
+        credentialManager.save(credentials: credentials)
+        var user = User(credential: credentials, clientName: self.keycloakConfig.clientID)
```
wei-lee: use `let`

modules/auth/authenticator/OIDCAuthenticator.swift
```diff
@@ -2,27 +2,91 @@
...
+        credentialManager.save(credentials: credentials)
+        onCompleted(User(credential: credentials, clientName: self.keycloakConfig.clientID), nil)
+    }
+    
+    func authFailure(error: Error?, onCompleted: @escaping (User?, Error?) -> Void) {
+        credentialManager.clear()
+        onCompleted(nil, error)
+    }
+    
+    public func resumeAuth(url: URL) -> Bool {
```
wei-lee: can we add some documents for the public methods

PR #33, opened 2018-03-05
Implement the User class

wei-lee:
### Motivation

https://issues.jboss.org/browse/AEROGEAR-2090

### Description

* Refactoring the `ConfigValue` class and make it easier to use
* Implement the `User` class and add tests for the auth module

### Progress

- [x] Refactoring `ConfigValue` class to `JSONValue` and add unit test
- [x] Implement the `User` class in the auth module and add unit test

### Additional Notes

The `ConfigValue` class was an enum type and the current implementation doesn't allow extracting the raw value. However, the type of the raw value is unknown, it could be string, bool, int or double. To support it, we have a few options:

1. Use something like `JSONObject`. Unfortunately, there is so such thing in Swift.
2. Use a Map with `Any` type for the value. But `Any` type doesn't conform to the `Codable` protocol which we have to use in order to using JSONCoder/JSONDecoder.

In the end, we have to create our own class (`JSONValue`)  that can represent the raw value. The class provides a few methods to allow extra the raw value and convert it to the right type automatically.

ping @rachael-oregan  This should unblock you

ping @wtrocki @danielpassos for review.


wtrocki:
+100 on `MobileService` implementation. Thanks for increasing code coverage.

PR #34, opened 2018-03-05
WIP AEROGEAR-2231 - Implement KeycloakConfig and AuthenticationConfig

rachael-oregan:
### Motivation

JIRA: https://issues.jboss.org/browse/AEROGEAR-2231

### Description

Implement KeycloakConfig and AuthenticationConfig

### Progress

- [x] KeycloakConfig implementation
- [x] AuthenticationConfig implementation


modules/auth/config/AuthenticationConfig.swift
```diff
@@ -7,4 +7,18 @@ import Foundation
...
+    init(redirectURL: String, minTimeBetweenJwksRequests: Int = 24*60) {
+        self.redirectURL = URL(string: redirectURL)!
+        self.minTimeBetweenJwksRequests = minTimeBetweenJwksRequests
+    }
+    
+    public func getRedirectURL() -> URL {
+        return redirectURL
+    }
+    
+    public func getMinTimeBetweenJwksRequests() -> Int {
```
wei-lee: it's ok to have those getters, in Swift, it is generally unnecessary to have getters. The properties here are constants so they are already readonly, which means we can just allow access to those properties to get the values. 

For variable properties, you do use something like [this](https://stackoverflow.com/questions/41273035/read-only-properties) to allow read-only properties.

modules/auth/config/KeycloakConfig.swift
```diff
@@ -22,9 +44,44 @@ class KeycloakConfig {
...
+    
+    // Get the URL where JWKs can be retrieved
+    public func getJwksUrl() -> String {
+        return "\(baseUrl)/certs"
+    }
+    
+    // Get the JWK Issuer
+    public func getIssuer() -> String {
+        return String(format: "%@/realms/%@", getHostUrl(), getRealmName())
     }
```
wei-lee: for those fields, it's better to use [Computed Properties](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/Properties.html#//apple_ref/doc/uid/TP40014097-CH14-ID254)

rachael-oregan:
@wei-lee mind having another look please

example/AeroGearSdkExampleTests/auth/authenticator/OIDCAuthenticatorTest.swift
```diff
@@ -39,7 +39,7 @@ class OIDCAuthenticatorTest: XCTestCase {
         // Put setup code here. This method is called before the invocation of each test method in the class.
         let serviceConfig = ServiceConfig()
         keycloakConfig = KeycloakConfig(serviceConfig)
-        authConfig = AuthenticationConfig(redirectURL: URL(string: "com.aerogear.mobile.test://calback")!)
+        authConfig = AuthenticationConfig.init(redirectURL: "com.aerogear.mobile.test://calback")
```
wei-lee: do we need to explicitly call `init` here? I thought it should be the constructor?

modules/auth/config/AuthenticationConfig.swift
```diff
@@ -7,4 +7,10 @@ import Foundation
 /** Configurations for the authentication service */
 struct AuthenticationConfig {
```
wei-lee: can we add some documentation?

modules/auth/config/KeycloakConfig.swift
```diff
@@ -22,9 +40,59 @@ class KeycloakConfig {
...
+    
+    // Get the client id string
+    var clientID: String {
+        get {
+            return clientId
+        }
+    }
+    
+    // Get the logout URL string
+    public func getLogoutUrl(idToken: String, redirectUri: URL) -> String {
```
wei-lee: make this computed property as well. Don't think we need to treat this differently.

modules/auth/config/KeycloakConfig.swift
```diff
@@ -22,9 +40,59 @@ class KeycloakConfig {
...
+    
+    // Get the URL string of the Keycloak service
+    var hostUrl: String {
+        get {
+            return serverUrl
+        }
+    }
+    
+    // Get the realm name of the Keycloak service
+    var realmName: String {
```
aidenkeating: Can we remove the explicit `get`, I think it's implied unless also doing a `set`? Should be a bit cleaner:

```swift
var realmName: String {
    return realmId
}
```

rachael-oregan:
@wei-lee @aidenkeating mind having another look please

modules/auth/config/AuthenticationConfig.swift
```diff
@@ -4,7 +4,22 @@
...
+    let minTimeBetweenJwksRequests: Int
+    
+    /**
+     Initialises the authentication configuration
+     
+     - parameters:
+        - redirectURL: the redirect URL for the developers app
+        - minTimeBetweenJwksRequests: The minimum time, in minutes, between Json web key set requests. Default value is 1400 (1 day)
+     */
+    init(redirectURL: String, minTimeBetweenJwksRequests: Int = 24*60) {
```
aidenkeating: Should `minTimeBetweenJwksRequests ` be a `UInt`. Don't think we want to allow negative integers here?

modules/auth/config/KeycloakConfig.swift
```diff
@@ -1,19 +1,54 @@
...
 //  AuthConfig.swift
 //  AGSAuth
-
 import AGSCore
 import Foundation
 
-/** All the configurations related to Keycloak */
+/**
+ All the configurations related to Keycloak
+ */
```
aidenkeating: We need tests for this

rachael-oregan: Would unit tests for getters be of much value?

modules/auth/authenticator/OIDCAuthenticator.swift
```diff
@@ -33,7 +33,8 @@ public class OIDCAuthenticator: Authenticator {
        - onCompleted: a block function that will be invoked when the logout is completed.
      */
     public func logout(currentUser: User, onCompleted: @escaping (Error?) -> Void) {
-        let logoutUrl = keycloakConfig.getLogoutUrl(idToken: currentUser.identityToken)
+        keycloakConfig.logoutURL = currentUser.identityToken
+        let logoutUrl = keycloakConfig.logoutURL
```
aidenkeating: Should we be standardizing on `Url` v `URL`? Swift seems to suggest `URL` is the way to go here.

rachael-oregan: I'll update this

modules/auth/config/KeycloakConfig.swift
aidenkeating: For some no there's not much, but others like `logoutURL` completely modify what is provided to it right? I think it'd be good to assert the transformation for regression testing

```swift
myKeycloakConfig.logoutURL = "abc123"
print(myKeycloakConfig.logoutURL) // This won't print abc123 right? It'll be the full URL
```

rachael-oregan: Ya ok fair enough, I'll add tests for this

modules/auth/config/KeycloakConfig.swift
```diff
@@ -22,9 +57,68 @@ class KeycloakConfig {
...
+     
+     Set - sets the logout URL String. Expects identity token as a parameter
+     */
+    var logoutURL: String {
+        get {
+            return logoutUrl
+        }
+        set(idToken) {
+            logoutUrl = String(format: logoutUrlTemplate, baseUrl, tokenHintFragment, idToken, redirectFragment, authConfig.redirectURL.absoluteString)
+        }
```
wei-lee: -1 for using property for the logout url. The setter is setting the `idToken`, which is against most developer's expectation. Setter should just set the value for logoutUrl.

I suggest change this back to a function. Maybe call it `buildLogoutUrl` so that it doesn't look like a getter function.

rachael-oregan: Ok I'll change it back to a function.

Does it make more sense to just pass the redirectURL into this `buildLogoutUrl` function instead of passing the authentication config into the keycloak configuration on init?

wei-lee: Maybe. But lets leave it there for now as in the future there might be more configuration options are added to it.

rachael-oregan:
@wei-lee @aidenkeating I think I've addressed your comments

modules/auth/config/KeycloakConfig.swift
```diff
@@ -22,9 +57,66 @@ class KeycloakConfig {
...
+    var clientID: String {
+        return clientId
+    }
+    
+    /**
+     Constructs the logout URL
+     
+     - parameters:
+        - idToken: the identity token
+        - redirectURI: the redirect uri
```
wei-lee: this parameter shouldn't be here.

PR #35, opened 2018-03-05
AEROGEAR-2092 Add OIDCCredentials implementation

aidenkeating:
### Motivation

We need a way to store and manage credentials.

JIRA: [AEROGEAR-2092](https://issues.jboss.org/browse/AEROGEAR-2092)

### Description

Create an OIDCCredentials object which is backed by an OIDAuthState.

Create a CredentialManager which takes the OIDAuthState from the OIDCCredentials and stores it using a Keychain storage library.

### Progress

- [x] OIDCCredentials
- [x] CredentialManager
- [x] OIDCCredentialsTest
- [x] CredentialManagerTest
- [x] Make OIDCCredentials NSCoding compliant (so the CredentialManager doesn't need to know about authState)

### DIFF
--- a/example/AeroGearSdkExampleTests/auth/credentials/OIDCCredentialsTest.swift
+++ b/example/AeroGearSdkExampleTests/auth/credentials/OIDCCredentialsTest.swift
@@ -84,11 +84,13 @@ internal class OIDCCredentialsTest: XCTestCase {
         XCTAssertFalse(unauthorizedCredentials.isAuthorized())
     }
 
-    /// Create an OIDCCredentials instance backed by an OIDAuthState with the provided parameters for a token response.
-    /// This allows for mocking most of the important values that are used by OIDCCredentials.
-    ///
-    /// - Parameter parameters: The parameters to mock being provided by a token response e.g. access_token, expires_in.
-    /// - Returns: An OIDCCredentials instance backed by a mocked AuthState using the parameters provided.
+    /**
+     Create an OIDCCredentials instance backed by an OIDAuthState with the provided parameters for a token response.
+     This allows for mocking most of the important values that are used by OIDCCredentials.
+ 
+     - Parameter parameters: The parameters to mock being provided by a token response e.g. access_token, expires_in.
+     - Returns: An OIDCCredentials instance backed by a mocked AuthState using the parameters provided.
+    */
     internal static func buildCredentialsWithParameters(parameters: [String: NSObject & NSCopying]) -> OIDCCredentials {
         let testUrl = URL(string: "https://example.example")!
         let testServiceConfig = OIDServiceConfiguration(authorizationEndpoint: testUrl, tokenEndpoint: testUrl)
--- a/example/AeroGearSdkExampleTests/auth/credentials/OIDCCredentialsTest.swift
+++ b/example/AeroGearSdkExampleTests/auth/credentials/OIDCCredentialsTest.swift
@@ -52,15 +52,15 @@ internal class OIDCCredentialsTest: XCTestCase {
     }
 
     func testGetAccessToken() {
-        XCTAssert(testCredentials?.getAccessToken() == OIDCCredentialsTest.paramAccessTokenVal)
+        XCTAssertEqual(testCredentials?.getAccessToken(), OIDCCredentialsTest.paramAccessTokenVal)
     }
 
     func testGetIdentityToken() {
-        XCTAssert(testCredentials?.getIdentitityToken() == OIDCCredentialsTest.paramIdTokenVal)
+        XCTAssertEqual(testCredentials?.getIdentitityToken(), OIDCCredentialsTest.paramIdTokenVal)
     }
 
     func testGetRefreshToken() {
-        XCTAssert(testCredentials?.getRefreshToken() == OIDCCredentialsTest.paramRefreshTokenVal)
+        XCTAssertEqual(testCredentials?.getRefreshToken(), OIDCCredentialsTest.paramRefreshTokenVal)
     }
 
     func testIsExpired() {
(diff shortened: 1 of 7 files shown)


modules/auth/credentials/OIDCCredentials.swift
```diff
@@ -5,14 +5,47 @@
 import AppAuth
 import Foundation
 
-class OIDCCredentials {
+/// A class to represent the OpenID Connect state for an entity.
```
wtrocki: let's use `/**` format

modules/auth/credentials/OIDCCredentials.swift
```diff
@@ -5,14 +5,47 @@
...
+/// This is backed by a standard OIDC token.
+public class OIDCCredentials {
     let authState: OIDAuthState
 
-    init(state: OIDAuthState) {
-        authState = state
+    public init(state: OIDAuthState) {
+        self.authState = state
```
wtrocki: That will break lint. Default rules are preventing from using self.

modules/auth/agsauth.podspec
```diff
@@ -18,5 +18,6 @@ Pod::Spec.new do |s|
   s.dependency 'AGSCore'
   ## Add other dependencies if needed
   s.dependency 'AppAuth'
+  s.dependency 'SwiftKeychainWrapper'
```
wtrocki: We probably need to lock version.

wtrocki: We probably need to lock version as this third party library can be possible security threat if we will always get latest.

example/AeroGearSdkExampleTests/auth/credentials/OIDCCredentialsTest.swift
```diff
@@ -0,0 +1,123 @@
...
+        super.setUp()
+        testCredentials = OIDCCredentialsTest.buildCredentialsWithParameters(parameters: OIDCCredentialsTest.defaultParameters)
+    }
+    
+    override func tearDown() {
+        super.tearDown()
+    }
+    
+    func testGetAccessToken() {
+        XCTAssert(testCredentials?.getAccessToken() == OIDCCredentialsTest.paramAccessTokenVal)
```
wtrocki: `XCTAssertEqual`?

wtrocki:
@aidenkeating Awesome changes! You need to run `./scripts/develop.sh` to fix some formatting/style.

modules/auth/agsauth.podspec
wtrocki: Actually library seems to be just changing the api, but delivering no extra capabilities. It's up to you but it may make sense to just use plain apple api to simplify IOS versions support etc.

modules/auth/credentials/CredentialsManager.swift
```diff
@@ -3,23 +3,42 @@
...
     func save(credentials: OIDCCredentials)
     func clear()
 }
 
 /** Persist/load the OIDCCredentials */
-class CredentialsManager: CredentialManagerProtocol {
-    func load() -> OIDCCredentials? {
-        return nil
+public class CredentialsManager: CredentialManagerProtocol {
+    let AUTH_STATE_KEY = "authState"
```
wei-lee: use reverse domain style key to ensure uniqueness of the key

modules/auth/credentials/OIDCCredentials.swift
```diff
@@ -5,14 +5,47 @@
...
+    public func isExpired() -> Bool {
+        return (authState.lastTokenResponse?.accessTokenExpirationDate)! < Date()
+    }
+
+    /// If the credential is in an authorized state.
+    ///
+    /// - Returns: true if the credential is authorized.
+    public func isAuthorized() -> Bool {
+        return authState.isAuthorized
     }
```
wei-lee: So in swift, when you need to get a value from a class, if there is no parameter required to get the value, you normally have 2 options:

1. Use [computed properties](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/Properties.html#//apple_ref/doc/uid/TP40014097-CH14-ID254), or
2. Use a function that takes no parameters and return a value

Functionality-wise, they are the same,  but there is subtle differences semantically. 

Using computed properties means the value is a property of the class, while using functions means the value is an method/action of the class.

For this class, I think most of the values returned by the getters are actually properties (they are properties on the AuthState class as well). So from that point of view, it's better to use computed properties instead. It is also more concise to write.

WDYT?

aidenkeating: Yes this makes a lot of sense. Good point.

modules/auth/agsauth.podspec
wei-lee: @wtrocki The library makes it a lot simpler to set/read/delete data in the Keychain. The Keychain API isn't the easiest one to use in the world. What is the downside of user the lib?

wtrocki: @wei-lee Not a problem. Just better to specify the version range for that.

wei-lee: @wtrocki agreed.

modules/auth/credentials/CredentialsManager.swift
```diff
@@ -2,24 +2,46 @@
...
 //  AGSAuth
 
+import AppAuth
 import Foundation
+import Security
+import SwiftKeychainWrapper
 
-protocol CredentialManagerProtocol {
+public protocol CredentialManagerProtocol {
     func load() -> OIDCCredentials?
```
wtrocki: [Optional] Comments are usually more important on the protocol.

aidenkeating:
@wtrocki @wei-lee Mind taking another look, just ensured the OIDCCredentials itself is encodable instead of having the CredentialManager be aware of stuff it shouldn't be.

PR #36, opened 2018-03-06
Ag 2089 implement auth

wei-lee:
### Motivation

https://issues.jboss.org/browse/AEROGEAR-2089

### Description

Wire up the login/logout/currentUser functions in the AgsAuth module. 

While working on this, I also discovered that we are missing a function to parse the access token value and convert it to User model, so I implemented it as well.

### Progress

- [x] Make the required code changes
- [x] Add unit tests
- [x] Update docs.

ping @aidenkeating @rachael-oregan @ziccardi


example/AeroGearSdkExampleTests/auth/config/KeycloakConfigTests.swift
```diff
@@ -6,21 +6,39 @@
...
 @testable import AGSAuth
-import AGSCore
+@testable import AGSCore
 import Foundation
 import XCTest
 
 class KeycloakConfigTests: XCTestCase {
     
-    var configService: ServiceConfig?
+    let mobileServiceData =
```
wtrocki: [Optional][Trivial] We could extract that to separate mock and reuse it for tests.

example/AeroGearSdkExampleTests/auth/config/KeycloakConfigTests.swift
```diff
@@ -6,21 +6,39 @@
...
+    """
+    {
+      "id": "keycloak",
+      "name": "keycloak",
+      "type": "keycloak",
+      "url": "https://www.mocky.io/v2/5a6b59fb31000088191b8ac6",
```
wtrocki: [Optional] Change to `https://keycloak-myproject.192.168.64.74.nip.io/auth` to not cause confusion.

modules/auth/AgsAuth.swift
```diff
@@ -2,32 +2,133 @@ import AGSCore
 import Foundation
 
 /**
- AeroGear Auth Service
+ AeroGear auth module.
+ 
+ It will allow users to perform login/logout actions against an Keycloak service provisioned by the AeroGear mobile service on OpenShift.
```
wtrocki: [Optinal] We can assume that is done in comment`It will allow` => `Allows`

modules/auth/AgsAuth.swift
```diff
@@ -2,32 +2,133 @@ import AGSCore
...
+ 
+ You can use the module like this:
```
wtrocki: - example:

modules/auth/AgsAuth.swift
```diff
@@ -2,32 +2,133 @@ import AGSCore
...
-    let credentialManager: CredentialsManager
-    let keycloakConfig: KeycloakConfig
-    let authServiceConfig: AuthenticationConfig
-
-    init(serviceConfig: ServiceConfig, authConfig: AuthenticationConfig) {
-        keycloakConfig = KeycloakConfig(serviceConfig, authConfig)
-        authServiceConfig = authConfig
+    private static let serviceId = "keycloak"
+
+    public static let instance = AgsAuth(AgsCore.instance.getConfiguration(serviceId)!)
```
wtrocki: [Important] We could rename that to `default` or make init method private.

modules/auth/AgsAuth.swift
```diff
@@ -2,32 +2,133 @@ import AGSCore
...
+
+    /** Errors generated by the auth module */
+    enum Errors: Error {
```
wtrocki: [Optional] Extract outside the class.

modules/auth/User.swift
```diff
@@ -49,23 +48,182 @@ struct  UserRole: Hashable {
...
+    private let name: String?
+    private let preferredName: String?
+    private let realmAccess: AccessRoles?
+    private let resourceAccess: [String: AccessRoles]?
+
+    let firstName: String?
+    let lastName: String?
+    let email: String?
+
+    enum CodingKeys: String, CodingKey {
```
wtrocki: Note: this exposes only default user attributes - ignoring http://www.keycloak.org/docs/3.0/server_development/topics/custom-attributes.html

modules/auth/User.swift
```diff
@@ -49,23 +48,182 @@ struct  UserRole: Hashable {
...
+ Describe the structure of user profile returned by Keycloak
+ */
+struct KeycloakUserProfile: Codable {
+
+    /** Internal structure for AccessRoles. The JSON object contains other fields but we only interested in "roles" */
+    struct AccessRoles: Codable {
+        let roles: [String]?
+    }
+
+    private let name: String?
```
wtrocki: Note: this exposes only default user attributes - ignoring http://www.keycloak.org/docs/3.0/server_development/topics/custom-attributes.html

Not an issue. Just mentioning for awareness. Not sure if we need to allow users to retrieve their custom stuff. 
If you make this open users could extend it if they need it.

modules/auth/User.swift
```diff
@@ -49,23 +48,182 @@ struct  UserRole: Hashable {
     }
 }
 
+/**
+ Describe the structure of user profile returned by Keycloak
+ */
+struct KeycloakUserProfile: Codable {
```
wtrocki: Let's make this public.

modules/auth/User.swift
```diff
@@ -49,23 +48,182 @@ struct  UserRole: Hashable {
...
+     - returns:
+     An array of the role names
+    */
+    func getClientRoles(_ clientName: String) -> [String] {
+        let clientRolesArr = [String]()
+        guard  let resourceData = resourceAccess,
+            let clientData = resourceData[clientName],
+            let clientRoles = clientData.roles
+        else {
+            return clientRolesArr
```
wtrocki: return [String]()

modules/auth/User.swift
```diff
@@ -49,23 +48,182 @@ struct  UserRole: Hashable {
...
+
+    /**
+     Return the client roles of the user
+     - parameters:
+         - clientName: the name of the client
+     - returns:
+     An array of the role names
+    */
+    func getClientRoles(_ clientName: String) -> [String] {
+        let clientRolesArr = [String]()
```
wtrocki: Not used

modules/auth/User.swift
```diff
@@ -49,23 +48,182 @@ struct  UserRole: Hashable {
...
+         - credential: the OIDCCredentials
+         - clientName: the name of the Keycloak client
+     */
+    internal init?(credential: OIDCCredentials, clientName: String) {
+        guard let token = credential.accessToken,
+            let jwt = try? Jwt.decode(token)
+        else {
+            return nil
+        }
+        let payload = jwt.payload
```
wtrocki: Note: it's not recomended to encode large sets of data into JWT payload.

wtrocki: [Note] It is not recommended to encode large sets of data into JWT payload.

wtrocki: Edit: Actually that comes from keycloak. Sorry.

modules/auth/User.swift
```diff
@@ -49,23 +48,182 @@ struct  UserRole: Hashable {
...
 }
 
+/**
+ Describe the structure of user profile returned by Keycloak
+ */
+struct KeycloakUserProfile: Codable {
+
+    /** Internal structure for AccessRoles. The JSON object contains other fields but we only interested in "roles" */
+    struct AccessRoles: Codable {
+        let roles: [String]?
```
wtrocki: `[String]()` will not need optional

modules/auth/User.swift
```diff
@@ -49,23 +48,182 @@ struct  UserRole: Hashable {
...
+     Check if the user has a realm fole
+     - parameters:
+         - roleName: Name of the role
+     - returns:
+     If user has the role
+     */
     public func hasRealmRole(_ roleName: String) -> Bool {
         let roleToFind = UserRole(nameSpace: nil, roleName: roleName)
-        return roles.contains(roleToFind)
+        return roles!.contains(roleToFind)
```
wtrocki: [Optional] `!` could be avoided.

modules/auth/jwt/Jwt.swift
```diff
@@ -0,0 +1,47 @@
...
+        .replacingOccurrences(of: "-", with: "+")
+        .replacingOccurrences(of: "_", with: "/")
+
+    if base64.count % 4 != 0 {
+        base64.append(String(repeating: "=", count: 4 - base64.count % 4))
+    }
+    return Data(base64Encoded: base64)
+}
+
+//TODO: use a proper library and add verification
```
wtrocki: [Question] Do we want to use use JWT secret?

wtrocki:
@wei-lee Great changes. Let me know if you are happy for me to play with that in demo application and verify the work. Great to see Auth SDK getting into the right shape. I have made lots of optional comments. Feel free to remove them to clear things up.

modules/auth/AgsAuth.swift
wei-lee: this is a much better document than Apple's...

modules/auth/AgsAuth.swift
wei-lee: Normally if you use `default`, it normally implies there could be ones that are not `default`. But in this case we don't want to users to create their own. It is also the same as the core module.

By default the init method is internal, which means it can not be called outside this module. It also makes it possible to create new instance of the service. `private` doesn't allow to use it in tests.

modules/auth/AgsAuth.swift
wei-lee: I think using `AgsAuth.Error.someError` is ok? It also makes it clear where the errors belongs to?

modules/auth/User.swift
wei-lee: yes, I am aware of that. We are not supporting custom attributes yet. It will be something to add in the future.

modules/auth/User.swift
wei-lee: for now it's not supposed to be used by developers so we want to keep it internal.

modules/auth/jwt/Jwt.swift
wei-lee: Keycloak uses public/private key

wei-lee:
@wtrocki I addressed most of your comments, mind taking a look again?

PR #42, opened 2018-03-09
Push SDK

wtrocki:
### Motivation

Add push SDK

screen shot 2018-03-15 at 4 18 48 pm


modules/push/DeviceRegistration.swift
```diff
@@ -0,0 +1,240 @@
...
+    @objc var overrrideProperties: [String: String]?
+    
+    /**
+    An initializer method to instantiate an DeviceRegistration object.
+    
+    :param: serverURL the URL of the AeroGear Push server.
+    
+    :returns: the DeviceRegistration object.
+    */
+    @objc public init(serverURL: URL) {
```
wtrocki: @danielpassos Do you think it makes sense to depend on core here or just pass the url. It looks like this could work pretty much as independent upstream library.

wtrocki:
Tested with latest version of UPS docker image.

modules/push/DeviceRegistration.swift
danielpassos: I'd like to have it working easily for people using Openshift and also a standard alone UPS.

Wdyt?

wtrocki: It will depend on the outcome of the conversation here: https://github.com/aerogear/proposals/pull/31

wtrocki:
@danielpassos Work mostly done. I'm on the testing phase. 
Going to add some minor improvements to UI etc.
Any suggestions welcome.

example/AeroGearSdkExample/push/PushHelper.swift
```diff
@@ -0,0 +1,91 @@
...
+* Contains set of methods used to handle push configuration and messaging.
+*/
+public class PushHelper {
+    
+    /**
+      Called when push message was recieved.
+      Presents dialog with message
+    */
+    public func onPushMessageRecieved(_ userInfo: [AnyHashable : Any], _ fetchCompletionHandler: (UIBackgroundFetchResult) -> Void) {
+        AgsCore.logger.info("Push message recieved: \(userInfo)")
```
danielpassos: Typo: received

example/AeroGearSdkExample/push/PushHelper.swift
```diff
@@ -0,0 +1,91 @@
...
+                
+                let currentDevice = UIDevice()
+                clientDevice.alias = currentDevice.name
+                clientDevice.operatingSystem = currentDevice.systemName
+                clientDevice.osVersion = currentDevice.systemVersion
+                clientDevice.deviceType = currentDevice.model
+        }, success: {
+            AgsCore.logger.info("Successfully registered to unified push server")
+        }, failure: { (error: Error!) in
+            AgsCore.logger.error("Failure to register for notifications on UPS server: \(error)")
```
danielpassos: If you have used `unified push server` on success let's use the same string here instead of UPS

example/AeroGearSdkExample/push/PushHelper.swift
```diff
@@ -0,0 +1,91 @@
...
+    }
+    
+    /**
+     Initial setup for push notifications
+     - registration for remote configuration
+     - granting permissions from users
+    */
+    public func setupPush() {
+        UIApplication.shared.registerForRemoteNotifications()
+        if #available(iOS 10.0, *) {
```
danielpassos: Nice trick ;)

modules/push/ClientDeviceInformationImpl.swift
```diff
@@ -0,0 +1,43 @@
...
+ 
+    override init() {
+        super.init()        
+    }
+    
+    @objc func extractValues() -> [String: AnyObject] {
+        var jsonObject =  [String: AnyObject]()
+        jsonObject["deviceToken"] = convertToString(deviceToken) as AnyObject?
+        jsonObject["alias"] = alias as AnyObject?
+        jsonObject["categories"] = categories as AnyObject?
```
danielpassos: `alias` and `categories` should not be on JSON file.

modules/push/DeviceRegistration.swift
```diff
@@ -0,0 +1,115 @@
...
+                failure(error as NSError!)
+                return
+            }
+            self.saveClientDeviceInformation(clientInfoObject, serverURLGuard)
+            success();
+        });
+    }
+    
+    // Storing data in UserDefaults
+    fileprivate func saveClientDeviceInformation(_ clientInfoObject: ClientDeviceInformation, _ serverURLGuard: URL) {
```
danielpassos: Why do we need to "store" this info?

modules/push/ClientDeviceInformationImpl.swift
wtrocki: This is send on registration. Upstream does that as well.

modules/push/DeviceRegistration.swift
wtrocki: So this is to recall and change flavour for client  later. Got that from upstream to basically know if application was registered

wtrocki: So this is to recall and change flavour for client  later. Got that from upstream to basically know if application was registered

modules/push/ClientDeviceInformationImpl.swift
danielpassos: Hrm, I can be wrong but IMO it's wrong. Alias, and Categories are runtime stuffs. For example, alias can be the user credentials and categories the things the user has interesting. This is a runtime stuffs there is no way to know that before (to add in the JSON) on in runtime. Is that makes sense?

modules/push/DeviceRegistration.swift
danielpassos: Hrm, I'm not against it but I think it's not necessary if we are not really using this information. I mean, dev should fire register everytime when app opened and should handle itself if it was successful register while the app is opened. Do you got my point?

wtrocki: Removed that

modules/push/ClientDeviceInformationImpl.swift
wtrocki: This is actually added at runtime in the app. I believe that we need to refactor and split this objects to two separate classes as in Android.

wtrocki:
@danielpassos Thank you! Better alignment in terms of config is really needed. I was heavily inspired by existing AeroGear IOS SDK's, but aproach in current Android SDK seems to be applying the best patterns and we should follow that!

example/AeroGearSdkExample/push/PushHelper.swift
```diff
@@ -0,0 +1,91 @@
...
+                // setup configuration
+                clientDevice.deviceToken = deviceToken
+                clientDevice.variantID = "01862a41-4d17-45d9-b0ab-5b5047d5d15d"
+                clientDevice.variantSecret = "d727538d-af49-4d60-b91d-dc37c1c2011a"
+                
+                let currentDevice = UIDevice()
+                clientDevice.alias = currentDevice.name
+                clientDevice.operatingSystem = currentDevice.systemName
+                clientDevice.osVersion = currentDevice.systemVersion
+                clientDevice.deviceType = currentDevice.model
```
wei-lee: do we really need developers to popular these info?

I believe we can load `variantID` and `variantSecret` from the configuration file. the device info can be populated by the SDK.

So developers really only need to pass in the `deviceToken`?

modules/push/ClientDeviceInformationImpl.swift
```diff
@@ -0,0 +1,43 @@
...
+        jsonObject["operatingSystem"] = operatingSystem as AnyObject?
+        jsonObject["osVersion"] = osVersion as AnyObject?
+        jsonObject["deviceType"] = deviceType as AnyObject?
+
+        return jsonObject;
```
wei-lee: is there any reason we are not using `JSONEncoder` here?

modules/push/ClientDeviceInformation.swift
```diff
@@ -0,0 +1,50 @@
+import Foundation
+
+/**
+ Represents the set of allowed device metadata.
+ */
+@objc 
+public protocol ClientDeviceInformation {
```
wei-lee: do we really need a protocol for the client device info? What about just a `struct`?

modules/push/DeviceRegistration.swift
```diff
@@ -0,0 +1,116 @@
...
+    var serverURL: URL!
+    var requestApi: AgsHttpRequestProtocol;
+    
+    /**
+     An initializer method to instantiate an DeviceRegistration object.
+    
+     - parameter serverURL: the URL of the AeroGear Push server.
+     - returns: the DeviceRegistration object.
+    */
+    @objc public init(_ serverURL: URL,_ requestApi: AgsHttpRequestProtocol) {
```
wei-lee: I think it's better to provide labels if there are multiple arguments

modules/push/DeviceRegistration.swift
```diff
@@ -0,0 +1,116 @@
...
+
+        let clientInfoObject = ClientDeviceInformationImpl()
+        clientInfoConfigurationBlock(clientInfoObject)
+
+        // deviceToken could be nil then retrieved it from local storage (from previous register).
+        // This is the use case when you update categories.
+        if clientInfoObject.deviceToken == nil {
+            clientInfoObject.deviceToken = UserDefaults.standard.object(forKey: DeviceRegistration.deviceTokenKey) as? Data
+        }
+        guard let serverURLGuard = self.serverURL else {
```
wei-lee: this can be moved right after the first `guard` check

modules/push/DeviceRegistration.swift
```diff
@@ -0,0 +1,116 @@
...
+        success:(() -> Void)!, failure:((NSError) -> Void)!) -> Void {
+
+        // can't proceed with no configuration block set
+        guard let clientInfoConfigurationBlock = clientInfo else {
+            failure(NSError(domain: DeviceRegistrationError.PushErrorDomain, code: 0, userInfo: [NSLocalizedDescriptionKey: "configuration block not set"]))
+            return
+        }
+
+        let clientInfoObject = ClientDeviceInformationImpl()
+        clientInfoConfigurationBlock(clientInfoObject)
```
wei-lee: not a big fan of allowing a function to change the state of a parameter. For readability, I think it's better for the block to return a new value of the modified `clientInfoObject`. 
Also not really sure why this is required.

wtrocki: Good catch. I kept this to be aligned with the code snippets that are printed in UPS, but it looks like we going to have OpenShift integration and this is no longer needed!

PR #43, opened 2018-03-12
[WIP] Allow empty JSON responses in HTTP module

aidenkeating:
### Motivation

Currently HTTP responses will result in an error if a responses body
is empty, even though the status code is 2xx.

### Description

This allows the HTTP check to succeed regardless of whether there is
a body to the response. This was discovered as part of the Keycloak
logout request.

### Progress

- [x] Add isHTTPResponseSuccess function


aidenkeating:
@wtrocki Closing this based on our conversation

wtrocki:
@aidenkeating As mentioned offline. We could also check if alamofire empty content error happened and this will resolve this case. However it feels that this will be kinda hack.

wtrocki:
This aproach works fine, but it will introduce unexpected behavior if api is not returning valid JSON.  This is also good example where wrappers do not fit to all use cases. Going to try couple cases and see if we can improve this.

Issue #44, opened 2018-03-12
HTTP requests fail when response body is empty due to JSON parsing


aidenkeating:
### Current behaviour
Currently when a request is made using the HTTP module it is assumed to have a JSON response. This causes a request in the Auth SDK to fail when the body of the response is empty. The HTTP status code is 200 though.

### Expected behaviour
The request should succeed when the response body is empty.

## Steps to reproduce
* Run the example app
* Login using auth
* Attempt to logout

wtrocki:
Logged issue to alamofire: https://github.com/Alamofire/Alamofire/issues/2466
Going to provide local workaround.

wtrocki:
Done

PR #45, opened 2018-03-12
Allow empty json responses

wtrocki:
### Motivation

https://github.com/aerogear/aerogear-ios-sdk/issues/44

## Progress

I have tried worked on top of raw status message check that was covering many cases where status code was ok but json were not parsable.

We have tested case to get raw response. 

## Status

Implemented error handling alternative: https://github.com/aerogear/aerogear-ios-sdk/pull/43

Issue upstream (however this may be expected behavior and we will get that rejected) https://github.com/Alamofire/Alamofire/issues/2466

## Notes

If we are not happy to get that wrapper in core we could do the same error handling in keycloak to detect if error is empty content and skip that for login method. 
@aidenkeating  Leaving up to you what will be more suitable aproach.

### DIFF
--- a/modules/core/http/AgsHttpRequest.swift
+++ b/modules/core/http/AgsHttpRequest.swift
@@ -20,22 +20,26 @@ public class AgsHttpRequest: AgsHttpRequestProtocol {
     public func get(_ url: String, params: [String: AnyObject]? = [:], headers: [String: String]? = [:],
                     _ handler: @escaping (Any?, Error?) -> Void) {
         Alamofire.request(url, parameters: params, headers: headers).responseJSON { (responseObject) -> Void in
-            if responseObject.result.isSuccess {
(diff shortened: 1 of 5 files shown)


modules/core/http/AgsHttpRequest.swift
```diff
@@ -16,50 +16,48 @@ public protocol AgsHttpRequestProtocol {
...
+        guard case let .failure(error) = response.result else {
+            handler(response.result.value, nil)
+            return
+        }
+        if let error = error as? AFError {
+            switch error {
+            case .responseSerializationFailed(let reason):
+                // Workaround. See: https://github.com/Alamofire/Alamofire/issues/2466
+                if case .inputDataNilOrZeroLength = reason  {
+                    handler([], nil)
```
wtrocki: Using dictionary here, but it could be also `nil`

PR #47, opened 2018-03-13
fix: Apply default style changes to the code

wtrocki:
### Motivation

Run lint and formatter (it looks like most of the code were not processed by formatter and linter)


modules/core/metrics/impl/DeviceMetrics.swift
```diff
@@ -14,7 +14,7 @@ public class DeviceMetrics: Metrics {
         return [
             "platform": "ios",
             "platformVersion": UIDevice.current.systemVersion,
-            "device": UIDevice.current.model
+            "device": UIDevice.current.model,
```
danielpassos: Should we have `,` in the last line?

modules/core/metrics/impl/AppMetrics.swift
```diff
@@ -20,7 +20,7 @@ public class AppMetrics: Metrics {
         return [
             "appId": self.appData.bundleId,
             "appVersion": self.appData.appVersion,
-            "sdkVersion": self.appData.sdkVersion
+            "sdkVersion": self.appData.sdkVersion,
```
danielpassos: Should we have `,` in the last line?

wtrocki: This is what lint does. Going to alter lint config to not insert that.

modules/core/metrics/AgsMetrics.swift
```diff
@@ -63,7 +63,7 @@ open class AgsMetrics: MetricsPublishable {
         return [
             "clientId": appData.clientId,
             "timestamp": Int(NSDate().timeIntervalSince1970 * 1000),
-            "data": payload
+            "data": payload,
```
danielpassos: Should we have `,` in the last line?

example/AeroGearSdkExample/auth/AuthDetailsViewController.swift
```diff
@@ -76,15 +76,15 @@ class AuthDetailsViewController: UIViewController, UITableViewDataSource, UITabl
...
+
     func showServiceNotFoundDialog() {
         let alertView = UIAlertController(title: "Logout Error", message: "Auth Service is not configured. Use AgsAuth.instance.configure", preferredStyle: UIAlertControllerStyle.alert)
         alertView.addAction(UIAlertAction(title: "Ok", style: .cancel, handler: { _ in }))
         present(alertView, animated: true, completion: nil)
     }
-    
+
     func onLogoutComplete(_: Error?) {
-        self.removeView()
```
danielpassos: I'm not sure if we should kill `self.` dunno about the patter Apple uses, but I like it

PR #49, opened 2018-03-14
Renaming podspec files so project is working on case-sensitive filesy…

wojta:
### Motivation

`pod install` is not working on case-sensitive filesystem. 
Causing 
```
[!] No podspec found for `AGSAuth` in `../modules/auth`
```
### Description

renaming files


wtrocki:
Thank you for contribution! Awesome change.

PR #50, opened 2018-03-14
fix: crash on auth config missing

wtrocki:
Prevent from crashing application when config is missing.
Enforce users to handle this error case.

### Progress

- [x] Lazy init instance
- [x] Unit tests


wtrocki:
@aidenkeating Happy to document some of this if this is not planned in near future.

aidenkeating:
@wtrocki 👀

example/AeroGearSdkExample/auth/AuthViewController.swift
```diff
@@ -14,7 +14,7 @@ class AuthViewController: UIViewController {
 
     override func viewDidLoad() {
         let authenticationConfig = AuthenticationConfig(redirectURL: "org.aerogear.mobile.example:/callback")
-        AgsAuth.instance.configure(authConfig: authenticationConfig)
+        try! AgsAuth.instance.configure(authConfig: authenticationConfig)
```
aidenkeating: Quick question: This will still cause a crash in the example app correct? We've moved the source of the crash out of the SDK?

modules/auth/AgsAuth.swift
```diff
@@ -19,57 +19,68 @@ import Foundation
...
-    public static let instance = AgsAuth(AgsCore.instance.getConfiguration(serviceId)!)
+    public static let instance: AgsAuth = {
+        let config = AgsCore.instance.getConfiguration(serviceId);
+        return AgsAuth(config)
+    }();
 
     /** Errors generated by the auth module */
     public enum Errors: Error {
-        /** Service is not configured */
+        /** Thrown when user auth configuration is missing */
```
aidenkeating: Is this a correct description? Looks like it's describing `missingConfig`.

modules/auth/AgsAuth.swift
```diff
@@ -19,57 +19,68 @@ import Foundation
...
         case serviceNotConfigured
+        /** Thrown when Auth config is missing in mobile-services.json */
+        case missingConfig
```
aidenkeating: Maybe `noServiceConfigurationFound` to be more descriptive?

modules/auth/AgsAuth.swift
```diff
@@ -19,57 +19,68 @@ import Foundation
...
-    private let serviceConfig: MobileService
-
     private var authenticator: Authenticator?
+    
+    private let serviceConfig: MobileService?
     private var keycloakConfig: KeycloakConfig?
-    private var authServiceConfig: AuthenticationConfig?
 
-    private var configured: Bool = false
+    private var configured: Bool {
```
aidenkeating: I think we can remove the explicit `get` here. I'm pretty sure it's implicit when there's no `set` defined.

```swift
private var configured: Bool {
    return keycloakConfig != nil
}
```

example/AeroGearSdkExample/auth/AuthViewController.swift
wtrocki: Yes. I made that on purpose.

aidenkeating: Yep, makes sense

modules/auth/AgsAuth.swift
wtrocki: Done

PR #53, opened 2018-03-21
AEROGEAR-2091 - JwksManager implementation

rachael-oregan:
### Motivation

JIRA: https://issues.jboss.org/browse/AEROGEAR-2091

### Description

This is just the JwksManager implementation used to load and save JWKS for a keycloak namespace

### Progress

- [x] Jwks manager class
- [x] unit tests
- [x] docs

### Additional Notes

In order to do the integrity checks we're trying to commit upstream to Keycloak to include the `x5t` field in the JWKS returned from Keycloak

### DIFF
--- a/example/AeroGearSdkExampleTests/auth/jwk/JwkManagerTests.swift
+++ b/example/AeroGearSdkExampleTests/auth/jwk/JwkManagerTests.swift
@@ -29,7 +29,7 @@ class JwkManagerTests: XCTestCase {
 
         // persist JWKS & date
         KeychainWrapper.standard.set(jwksString!, forKey: "myproject_jwks_content")
-        KeychainWrapper.standard.set(Date().description , forKey: "myproject_requested_date")
+        KeychainWrapper.standard.set(Date().timeIntervalSince1970, forKey: "myproject_requested_date")
     }
     
     override func tearDown() {
@@ -52,7 +52,7 @@ class JwkManagerTests: XCTestCase {
     
     func testFetchJwksIfNeededShouldRequest() {
         let day: TimeInterval = 60.0 * 60.0 * 24
-        let yesterday = Date().addingTimeInterval(-day).description
+        let yesterday = Date().addingTimeInterval(-day).timeIntervalSince1970
         KeychainWrapper.standard.set(yesterday, forKey: "myproject_requested_date")
         
         let needsFetch = jwksManagerToTest?.fetchJwksIfNeeded(keycloakConfig!, false)
--- a/example/AeroGearSdkExampleTests/auth/jwk/JwkManagerTests.swift
+++ b/example/AeroGearSdkExampleTests/auth/jwk/JwkManagerTests.swift
@@ -8,12 +8,11 @@ import SwiftKeychainWrapper
 import XCTest
 
 class JwkManagerTests: XCTestCase {
-    
     var http = MockHttpRequest()
     var keycloakConfig: KeycloakConfig?
     var authConfig: AuthenticationConfig?
     var jwksManagerToTest: JwksManager?
-    
+
     override func setUp() {
         super.setUp()
         // Put setup code here. This method is called before the invocation of each test method in the class.
@@ -21,61 +20,60 @@ class JwkManagerTests: XCTestCase {
         authConfig = AuthenticationConfig(redirectURL: "com.aerogear.mobile.test://calback")
         keycloakConfig = KeycloakConfig(serviceConfig, authConfig!)
         jwksManagerToTest = JwksManager(http, authConfig!)
-        
+
         // create JWKS
         let jwks = try? JSONSerialization.data(withJSONObject: getMockJwks(), options: [])
         let jwksString = String(data: jwks!, encoding: String.Encoding.ascii)
-        
 
         // persist JWKS & date
         KeychainWrapper.standard.set(jwksString!, forKey: "myproject_jwks_content")
         KeychainWrapper.standard.set(Date().timeIntervalSince1970, forKey: "myproject_requested_date")
     }
-    
+
     override func tearDown() {
         // Put teardown code here. This method is called after the invocation of each test method in the class.
         super.tearDown()
         KeychainWrapper.standard.removeObject(forKey: "myproject_jwks_content")
         KeychainWrapper.standard.removeObject(forKey: "myproject_requested_date")
     }
-    
+
     func testLoadExistingJwks() {
         let jwksResult = jwksManagerToTest?.load(keycloakConfig!)
-        let jwksKey: [String: String] = jwksResult!["keys"] as! [String : String]
+        let jwksKey: [String: String] = jwksResult!["keys"] as! [String: String]
         XCTAssertEqual(jwksKey["kid"], "adSoyXNAgQxV43eqHSiRZf6hN9ytvBNQyb2fFSdCTVM")
     }
-    
+
     func testFetchJwksIfNeededForceFetch() {
         let needsFetch = jwksManagerToTest?.fetchJwksIfNeeded(keycloakConfig!, true)
         XCTAssertTrue(needsFetch!)
     }
-    
+
     func testFetchJwksIfNeededShouldRequest() {
         let day: TimeInterval = 60.0 * 60.0 * 24
         let yesterday = Date().addingTimeInterval(-day).timeIntervalSince1970
         KeychainWrapper.standard.set(yesterday, forKey: "myproject_requested_date")
-        
+
         let needsFetch = jwksManagerToTest?.fetchJwksIfNeeded(keycloakConfig!, false)
         XCTAssertTrue(needsFetch!)
     }
-    
+
     func testFetchJwksWithCallback() {
         http.dataForGet = getMockJwks()
         var onCompletedCalled = false
-        jwksManagerToTest?.fetchJwks(keycloakConfig!, onCompleted: {(response, error) in
+        jwksManagerToTest?.fetchJwks(keycloakConfig!, onCompleted: { response, error in
             XCTAssertNil(error)
-            let jwksKey: [String: String] = response!["keys"] as! [String : String]
+            let jwksKey: [String: String] = response!["keys"] as! [String: String]
             XCTAssertEqual(jwksKey["kid"], "adSoyXNAgQxV43eqHSiRZf6hN9ytvBNQyb2fFSdCTVM")
             onCompletedCalled = true
         })
         XCTAssertTrue(onCompletedCalled)
     }
-    
+
     func testFetchJwksWithoutCallback() {
         jwksManagerToTest?.fetchJwks(keycloakConfig!)
         let persistedContent = KeychainWrapper.standard.string(forKey: "myproject_jwks_content")
         let jwksData = persistedContent?.data(using: .utf8)
-        let jwks = try? JSONSerialization.jsonObject(with: jwksData!, options: .mutableLeaves) as! [String : Any]
+        let jwks = try? JSONSerialization.jsonObject(with: jwksData!, options: .mutableLeaves) as! [String: Any]
         let jwksKeys = jwks!["keys"] as! [String: String]
         XCTAssertEqual(jwksKeys["kid"], "adSoyXNAgQxV43eqHSiRZf6hN9ytvBNQyb2fFSdCTVM")
     }
--- a/example/AeroGearSdkExampleTests/auth/jwk/JwkManagerTests.swift
+++ b/example/AeroGearSdkExampleTests/auth/jwk/JwkManagerTests.swift
@@ -1,0 +1,79 @@
+//
+//  JwkManagerTests.swift
+//  AeroGearSdkExampleTests
+//
+
+@testable import AGSAuth
+import SwiftKeychainWrapper
+import XCTest
+
+class JwkManagerTests: XCTestCase {
+    
+    var http = MockHttpRequest()
+    var keycloakConfig: KeycloakConfig?
+    var authConfig: AuthenticationConfig?
+    var jwksManagerToTest: JwksManager?
+    
+    override func setUp() {
+        super.setUp()
+        // Put setup code here. This method is called before the invocation of each test method in the class.
+        let serviceConfig = getMockKeycloakConfig()
+        authConfig = AuthenticationConfig(redirectURL: "com.aerogear.mobile.test://calback")
+        keycloakConfig = KeycloakConfig(serviceConfig, authConfig!)
+        jwksManagerToTest = JwksManager.init(http, authConfig!)
+    }
+    
+    override func tearDown() {
+        // Put teardown code here. This method is called after the invocation of each test method in the class.
+        super.tearDown()
+    }
+    
+    func testLoadExistingJwks() {
+        // create JWKS
+        let jwks = try? JSONSerialization.data(withJSONObject: getMockJwks(), options: [])
+        let jwksString = String(data: jwks!, encoding: String.Encoding.ascii)
+        
+        // persist JWKS & date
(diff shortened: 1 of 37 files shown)


example/AeroGearSdkExampleTests/auth/jwk/JwkManagerTests.swift
```diff
@@ -0,0 +1,79 @@
...
+    var authConfig: AuthenticationConfig?
+    var jwksManagerToTest: JwksManager?
+    
+    override func setUp() {
+        super.setUp()
+        // Put setup code here. This method is called before the invocation of each test method in the class.
+        let serviceConfig = getMockKeycloakConfig()
+        authConfig = AuthenticationConfig(redirectURL: "com.aerogear.mobile.test://calback")
+        keycloakConfig = KeycloakConfig(serviceConfig, authConfig!)
+        jwksManagerToTest = JwksManager.init(http, authConfig!)
```
wei-lee: I don't think you need to call `init` explicitly.

example/AeroGearSdkExampleTests/auth/jwk/JwkManagerTests.swift
```diff
@@ -0,0 +1,79 @@
...
+    }
+    
+    func testLoadExistingJwks() {
+        // create JWKS
+        let jwks = try? JSONSerialization.data(withJSONObject: getMockJwks(), options: [])
+        let jwksString = String(data: jwks!, encoding: String.Encoding.ascii)
+        
+        // persist JWKS & date
+        KeychainWrapper.standard.set(jwksString!, forKey: "myproject_jwks_content")
+        KeychainWrapper.standard.set(Date().description , forKey: "myproject_requested_date")
```
wei-lee: these can be created during setup, and should be removed on tear down

modules/auth/jwk/JwksManager.swift
```diff
@@ -0,0 +1,164 @@
...
+    
+    /**
+     Save the JWKS content for the given realm name locally using [SwiftKeychainWrapper](https://github.com/jrendel/SwiftKeychainWrapper).
+     
+     - parameters:
+        - keycloakRealmName: the realm associated with the JWKS
+        - jwks: the content of the JWKS
+    */
+    private func persistJwks(_ keycloakRealmName: String, _ jwks: String) {
+        let timeFetched = Date().description
```
wei-lee: not sure if this is a good way to save the date, and it may change depends on the locale setting of the device. 

I suggest we can just use [Date.timeIntervalSince1970](https://developer.apple.com/documentation/foundation/date/1779963-timeintervalsince1970) and the corresponding init method.

modules/auth/jwk/JwksManager.swift
```diff
@@ -0,0 +1,164 @@
...
+    public func fetchJwks(_ keycloakConfig: KeycloakConfig, onCompleted: (([String: Any]?, Error?) -> Void)? = nil) {
+        let jwksUrl = keycloakConfig.jwksUrl
+        http.get(jwksUrl, params: nil, headers: nil, {(response, error) -> Void in
+            if let error = error {
+                if (onCompleted != nil) {
+                    return onCompleted!(nil, error)
+                }
+                AgsCore.logger.error("Error fetching JWKS: \(error)")
+            }
+            if let response = response as? [String: Any] {
```
wei-lee: should be `else if`

modules/auth/jwk/JwksManager.swift
```diff
@@ -0,0 +1,164 @@
...
+                if let resData = try? JSONSerialization.data(withJSONObject: response, options: []),
```
wei-lee: the `if` statement above will convert the data into a map. If that works, I think it is enough and I don't think we need to convert the data to be JSON here.

rachael-oregan:
@wei-lee thanks for the review, I'll address your comments soon.  The plan was to contribute to Keycloak to include the `x5t` property in the response from the JWKS endpoint.  I thought this was the plan as discussed previously or has something changed? Maybe I misunderstood?

wei-lee:
@rachael-oregan it should be the `x5c` property. No nothing changed, I just want to check if this is still the plan

TommyJ1994:
@rachael-oregan Sorry for delay, here is an example of verifying JWT's using the exponent and modulus that are in JWKS endpoint.

Add to Podfile:
 `pod 'JSONWebToken', :git => 'https://github.com/kreactive/JSONWebToken.git', :branch => 'master'` 

Note: not sure if this is published or under what name - `pod JSONWebToken` references another library, hence why i'm referencing the repo directly.

Then:

`$ pod install` - might need to restart XCode.

```
import JSONWebToken
```

```
// The JWKS for the Secure-App Realm - https://keycloak.security.feedhenry.org/auth/realms/secure-app/protocol/openid-connect/certs
var jwksModulusString = "kr1fDOUrTZc1MnpY9brGiA7Cz6X1nX77pmrUEgnMq2mxU7ibSW0CAk5e5a4wkmLGYf8EyvaFPHT1fMrFmDK03oN8Q2anh-3e894cXBXazHzzaJD-Lz1HfOOZFeInkAasxWSo8KN1-Kg-1Z7QyrPLhfcbIwfH2Stabx-3lfEMtPGws7tqWg93piA8is1PwIV5_8k4CqLe7jNtUyYS4BKR07oBY6VVxXOKKQAQ3ToLN--sjfaXAjDuE1Go7iW9q7Yt6q9qu4JCX-k6IWu68y_H6cicLXwS1VXPMwFjDOj7cQZB7A3t4q0F-6NVL-t7UjrAAK_7V3lPB-rDwHO92iwlZw"
let jwksExponentString = "AQAB"

// Ensure the Base64 Modulus Representation Contains only legal Characters
jwksModulusString = jwksModulusString.replacingOccurrences(of: "_", with: "/").replacingOccurrences(of: "-", with: "+")

// Ensure the Base64 Modulus Representation has the Correct Length by Adding Padding if Necessary
jwksModulusString = jwksModulusString.padding(toLength: ((jwksModulusString.count + 3) / 4) * 4,
                  withPad: "=",
                  startingAt: 0)

// Convert the Modulus & Exponent from Strings to Data objects
let modulusData = Data(base64Encoded: jwksModulusString, options: .ignoreUnknownCharacters)
let exponentData = Data(base64Encoded: jwksExponentString, options: .ignoreUnknownCharacters)

do {
    // Attempt to Create a Public Key with the Modulus & Exponent
    let publicKey : RSAKey = try RSAKey.registerOrUpdateKey(modulus : modulusData!, exponent : exponentData!, tag : "publicKey")

    // A Raw JWT Access Token (Expires on 10/01/2031 @)
    let rawJWT = "eyJhbGciOiJSUzI1NiIsInR5cCIgOiAiSldUIiwia2lkIiA6ICJhZFNveVhOQWdReFY0M2VxSFNpUlpmNmhOOXl0dkJOUXliMmZGU2RDVFZNIn0.eyJqdGkiOiIwMGJhODRlYy01MTYzLTRhZDgtYWNmMC04ZTQ2Mzc4OWE0MmUiLCJleHAiOjE5NDg2MzI2NDgsIm5iZiI6MCwiaWF0IjoxNTE2NjMyNjQ4LCJpc3MiOiJodHRwczovL2tleWNsb2FrLnNlY3VyaXR5LmZlZWRoZW5yeS5vcmcvYXV0aC9yZWFsbXMvc2VjdXJlLWFwcCIsImF1ZCI6ImNsaWVudC1hcHAiLCJzdWIiOiJiMTYxN2UzOC0zODczLTRhNDctOGE2Yy01YjgyMmFkYTI3NWUiLCJ0eXAiOiJSZWZyZXNoIiwiYXpwIjoiY2xpZW50LWFwcCIsImF1dGhfdGltZSI6MCwic2Vzc2lvbl9zdGF0ZSI6ImMyNTVmMGFjLTgwOTEtNGM5Mi04ZjNhLTQ4ZmZiODgxYTcwYiIsInJlYWxtX2FjY2VzcyI6eyJyb2xlcyI6WyJtb2JpbGUtdXNlciJdfSwicmVzb3VyY2VfYWNjZXNzIjp7ImNsaWVudC1hcHAiOnsicm9sZXMiOlsiaW9zLWFjY2VzcyJdfSwiYWNjb3VudCI6eyJyb2xlcyI6WyJtYW5hZ2UtYWNjb3VudCIsIm1hbmFnZS1hY2NvdW50LWxpbmtzIiwidmlldy1wcm9maWxlIl19fX0.Df_B_o7xaVyF6yGzyjfZJLwF6J9jrmjlzkKGtWyBzblBFVs8HMNWhUen3QJIq_lJLEloXKMh0Hfd5Yo0iPbLlNFXb_qIuIUUDtHuQHEL5-LZbTW8WCvotbfVbO9hTHxXJXs5sWdL3wcS1T5PuC5l6mP1a1fppvkWPOlcRFWVy9f-aBSRiUNGoN_AlL2FUIeFeHKeyfw7F5EndQt9HRfoFr0vD1VuB7uM2UWhnscCwsFWZ2AMz_ZoiUQPjtUm-y3UMURFcAhIYHCw7xI1tPRVdMkMC8i87Qq9HJ90eflEDjqse6x67m48kT1wvI2qhoexPa-ALdWMrseWIdRqBmSUHw"

    // Create a new JWT Object using the JWT String
    let jwt : JSONWebToken = try JSONWebToken(string : rawJWT)

    // Create the RS256 Validator
    let validator = RegisteredClaimValidator.expiration &
        RegisteredClaimValidator.notBefore.optional & RSAPKCS1Verifier(key : publicKey, hashFunction: .sha256)

    // Get the Result of the Validation
    let validationResult = validator.validateToken(jwt)

    if(validationResult.isValid) {
        print("::: Authentic JWT")
    } else {
        print("::: Error Validating JWT. Cause: ", validationResult)
    }
} catch {
    print("::: RSA Public Key Could Not Be Created Using the Provided Modulus and Exponent:", error)
}
```

rachael-oregan:
@wei-lee I've addressed your requested changes. Let me know if these are ok/you have any more. Does it make sense to merge these changes even though they are not being used at the moment?

The plan is to implement the verification like @TommyJ1994 has described in his comment above (thanks Tom).  If the `JSONWebToken` gets published it should be fine.  Tom has opened an issue for it [here](https://github.com/kreactive/JSONWebToken/issues/9) so does it make sense to have these changes as separate PRs?

modules/auth/jwk/JwksManager.swift
```diff
@@ -0,0 +1,160 @@
...
+        let jwksUrl = keycloakConfig.jwksUrl
+        http.get(jwksUrl, params: nil, headers: nil, {(response, error) -> Void in
+            if let error = error {
+                if (onCompleted != nil) {
+                    return onCompleted!(nil, error)
+                }
+                AgsCore.logger.error("Error fetching JWKS: \(error)")
+            }
+            else if let response = response as? [String: Any] {
+                let resString = response.description
```
wei-lee: is it actually right to use `response.description`? I don't think we need to re-assign `response` here. We can do:

```
else if let _ = response as? [String: Any] {
    //continue use response here as string
}
```

modules/auth/jwk/JwksManager.swift
```diff
@@ -0,0 +1,160 @@
...
+    
+    /**
+     Save the JWKS content for the given realm name locally using [SwiftKeychainWrapper](https://github.com/jrendel/SwiftKeychainWrapper).
+     
+     - parameters:
+        - keycloakRealmName: the realm associated with the JWKS
+        - jwks: the content of the JWKS
+    */
+    private func persistJwks(_ keycloakRealmName: String, _ jwks: String) {
+        let timeFetched = Date.init(timeIntervalSince1970: Date().timeIntervalSince1970).description
```
wei-lee: why not just use `Date().timeIntervalSince1970` and save it?

modules/auth/jwk/JwksManager.swift
```diff
@@ -0,0 +1,160 @@
...
+     - returns: true if the request should be triggered, false otherwise
+    */
+    private func shouldRequestJwks(_ keycloakConfig: KeycloakConfig) -> Bool {
+        let namespace = keycloakConfig.realmName
+        let requestedDateEntryName = buildEntryNameForQuestedDate(namespace)
+
+        if let lastRequestDateString = KeychainWrapper.standard.string(forKey: requestedDateEntryName) {
+            let dateFormatter = DateFormatter()
+            dateFormatter.dateFormat = "yyyy-MM-dd HH:mm:ss ZZZ"
+            let lastRequestDate = dateFormatter.date(from: lastRequestDateString)
```
wei-lee: this is not required if `Date().timeIntervalSince1970` is used.

modules/auth/jwk/JwksManager.swift
rachael-oregan: `response` isn't a string here its `Any` so do you mean to try cast it to a String? Casting from `Any` to String fails.  Also casting `_` to String also fails.

Do you mean to use a struct that can be used to decode the structure or something?

wei-lee: right. So what is the actual type of `response`, a JSON object?

rachael-oregan: Yes a JSON object

rachael-oregan:
@wei-lee mind having another look please?

modules/auth/jwk/JwksManager.swift
```diff
@@ -0,0 +1,157 @@
...
+
+    */
+    public func fetchJwks(_ keycloakConfig: KeycloakConfig, onCompleted: (([String: Any]?, Error?) -> Void)? = nil) {
+        let jwksUrl = keycloakConfig.jwksUrl
+        http.get(jwksUrl, params: nil, headers: nil, { (response, error) -> Void in
+            if let error = error {
+                if onCompleted != nil {
+                    return onCompleted!(nil, error)
+                }
+                AgsCore.logger.error("Error fetching JWKS: \(error)")
```
wei-lee: the logger should happen before the `return` statement otherwise it may never be called

modules/auth/jwk/JwksManager.swift
```diff
@@ -0,0 +1,157 @@
...
+    */
+    private func shouldRequestJwks(_ keycloakConfig: KeycloakConfig) -> Bool {
+        let namespace = keycloakConfig.realmName
+        let requestedDateEntryName = buildEntryNameForQuestedDate(namespace)
+
+        if let lastRequestDate = KeychainWrapper.standard.double(forKey: requestedDateEntryName) {
+            let currentTime: Double = Date().timeIntervalSince1970
+            let duration = currentTime - lastRequestDate
+            let durationInMinutes = duration / 60
+            if abs(durationInMinutes) < Double(authConfig.minTimeBetweenJwksRequests) {
```
wei-lee: I don't think we need `abs` here?

rachael-oregan: Ya you're right. Nice catch

rachael-oregan:
@wei-lee mind having another look please

modules/auth/jwk/JwksManager.swift
```diff
@@ -0,0 +1,157 @@
...
+        - keycloakConfig: the keycloak service configuration that contains the keycloak server information to make requests to
+        - onCompleted: callback function to ebe invoked when the request is complete. Defaults to nil
+
+    */
+    public func fetchJwks(_ keycloakConfig: KeycloakConfig, onCompleted: (([String: Any]?, Error?) -> Void)? = nil) {
+        let jwksUrl = keycloakConfig.jwksUrl
+        http.get(jwksUrl, params: nil, headers: nil, { (response, error) -> Void in
+            if let error = error {
+                if onCompleted != nil {
+                    AgsCore.logger.error("Error fetching JWKS: \(error)")
```
wei-lee: is there any reason this is only logged when `onCompleted` is not nil? I think we should always log this.

PR #54, opened 2018-03-23
Jenkinsfile template for integration tests

psturc:
### Motivation

https://issues.jboss.org/browse/AEROGEAR-2207

### Description

* Jenkinsfile should contain configuration and commands for running only integration tests
* Integration test will be triggered only when PR contains label "test/integration"
* Implementation of integration test itself is currently blocked by https://issues.jboss.org/browse/AGIOS-675

### Progress

- [x] Jenkinsfile template for testing Mac OSX slave
- [x] Integration test


wtrocki:
@psturc Great stuff. Could we also disable reporting status back for non integration tests? I do not see label "integration test" - did you meant comment?

wtrocki:
Ignore my previous comment. My only recomendation is to use "test/integration" label. I created that label for you

ci/aerogear-app-metrics.yml
```diff
@@ -0,0 +1,143 @@
+---
```
wtrocki: -1 For having this app metrics template in IOS SDK.
It should be in target repository.

You can always `wget https://raw.github.com/...` to get recent version of that.

psturc: Yes, that's reasonable. I've created a PR https://github.com/aerogear/aerogear-app-metrics/pull/50

wtrocki:
@psturc Good to go!

psturc:
@wtrocki I'm waiting for https://issues.jboss.org/browse/AGIOS-675 to implement https://issues.jboss.org/browse/AEROGEAR-2207 and then put the command to run the integration test to the Jenkinsfile here.

psturc:
@wtrocki I think this PR is ready to be merged.

However, there's maybe one thing worth considering.
I was struggling with adding more tests, because they would have to be separated from regular unit tests (they would fail in 🔵 CI). And I did not find a nice way how to do the separation.
The only thing that worked, was renaming the tests to not start with name "test" so they are not triggered via `xcodebuild` or `fastlane` if you do not specify them, like [here](https://gist.github.com/psturc/a463ecc5e2a6bebcaaa964e0c75eb91d#file-metricsintegrationtest-swift-L5-L11). But that's not nice solution IMO.

Wdyt? Is there another way how we could achieve the separation of tests?

wtrocki:
@psturc Let's not hack things around. You need to create separate target for your tests and put tests into target. For example: http://rshankar.com/create-new-test-target-in-xcode/

As for current PR it's ok to merge and work on follow up PR's.

wtrocki:
@psturc Follow up tasks:

- [ ] Create separate target for functional tests
- [ ] Move tests to new target (or create new ones)
- [ ] Add info to contribution guide for integration tests (this should be technically adde to every bigger PR before merge.

PR #56, opened 2018-03-23
Aerogear 2116

rachael-oregan:
### Motivation

JIRA: https://issues.jboss.org/browse/AEROGEAR-2116

### Description

Reviewed and refactored the API docs for the Auth module.

### Progress

- [x] refactor typos and reword docs
- [x] refactor callbacks to include parameter labels so the labels show up in the docs and we can give a description for them
- [x] make docs one liners where appropriate
- [x] add full stops to descriptions and remove them from markup callouts like `parameters`, `returns` and `throws`


rachael-oregan:
@wei-lee @TommyJ1994 @aidenkeating @ziccardi if anyone has some time would you mind having a look thanks

modules/auth/AgsAuth.swift
```diff
@@ -77,24 +85,30 @@ open class AgsAuth {
      Perform user login action.
      
      - parameters:
-         - presentingViewController: the ViewController that initiates the login process
-         - onCompleted: callback function that will be invoked when the login is finished
+        - presentingViewController: the ViewController that initiates the login process
+        - onCompleted: callback function that will be invoked when the login is finished
+        - user: the user returned in the `onCompleted` callback function.  Will be nil if login failed
+        - error: the error returned in the `onCompleted` callback function. Will be nil if login was successful
```
wei-lee: is this the right place to list `user` and `error` here, given that they are the parameters for the callback function, but not the `login` function?

modules/auth/AgsAuth.swift
```diff
@@ -107,9 +121,13 @@ open class AgsAuth {
      Perform the logout action.
      
      - parameters:
-         - onCompleted: callback function that will be invoked when the logout process is completed
+        - onCompleted: callback function that will be invoked when the logout process has completed
+        - error: the `serviceNotConfigured` error returned in the `onCompleted` callback function
```
wei-lee: same comment as above

modules/auth/AgsAuth.swift
rachael-oregan: Ya its a bit odd but leaving out the parameter labels in the callback results in the following 

screen shot 2018-03-27 at 17 53 20 

Which to me looks worse, having `_ `and `no description`.  I tried looking into it more but I can't seem to indent or make it obvious that those parameters are for the callback function only

wei-lee: ah, ok. Never mind then. Better to have them listed.

Thanks.

PR #58, opened 2018-03-28
 📝 doc: add doc for SSO

wei-lee:
### Motivation

We need a document for SSO on iOS.

JIRA: [https://issues.jboss.org/browse/AEROGEAR-2129]

### Description

Add a documentation for SSO on iOS

### Progress

- [x] Add doc

### DIFF
--- a/docs/auth/sso.adoc
+++ b/docs/auth/sso.adoc
@@ -1,0 +1,37 @@
+== Single Sign-On (SSO)
+
+The Auth SDK uses OpenID Connect http://openid.net/specs/openid-connect-core-1_0.html#CodeFlowAuth[Auhtorization Code Flow] to achieve SSO via a mobile device browser. 
+
+'''
+
+=== Prerequisite's
+. A running AeroGear auth service (Keycloak)
+. Two iOS apps that have created and initialised with the Auth SDK module - link:README.adoc[AeroGear Services Auth SDK].
+
+'''
+
+=== Steps to achieve SSO via a device browser
+For this guide, let's assume that your two iOS apps are called **Email App** and **Messaging App**.
+
+==== Setup Keycloak
+
+. Ensure the "Remember Me" option is turned on. Click on *Realm Settings* section on the left hand side, and then click on *Login*. Turn on *Remember Me*.
+
+include::https://raw.githubusercontent.com/aerogear/aerogear-android-sdk/master/docs/getting-started/sso/keycloak.adoc[]
+
+==== Configuration File in the Mobile Apps
+
+From following the link:README.adoc[AeroGear Services Auth SDK] setup, there should be a configuration file, `mobile-services.json`, inside the iOS projects. It's important that the Keycloak configuration defined in this file matches the Keycloak client that the app intends to use.
+
+'''
+=== Perform SSO
+
+Now that your two iOS apps and Keycloak clients are setup, sign into the *Email App* with the user you have previously created. You should be redirected to your device browser which should have loaded the Keycloak login page.  
+
+Next, enter in your credentials, make sure check the *Remember Me* option and login. You have now been authenticated on the *Email App* via the browser.  
+
+Finally, try sign into the *Messaging App* and you should be automatically logged in as you have been previously authenticated in the *Email App* via the *_same_* browser.
+
+=== Known Issues
+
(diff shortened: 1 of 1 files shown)


wei-lee:
@rachael-oregan @TommyJ1994 mind reviewing this?

docs/auth/sso.adoc
```diff
@@ -0,0 +1,37 @@
...
+'''
+=== Perform SSO
+
+Now that your two iOS apps and Keycloak clients are setup, sign into the *Email App* with the user you have previously created. You should be redirected to your device browser which should have loaded the Keycloak login page.  
+
+Next, enter in your credentials, make sure check the *Remember Me* option and login. You have now been authenticated on the *Email App* via the browser.  
+
+Finally, try sign into the *Messaging App* and you should be automatically logged in as you have been previously authenticated in the *Email App* via the *_same_* browser.
+
+=== Known Issues
```
TommyJ1994: This issue is mainly related to our implementation of using the Safari inappbrowser rather than the Safari system browser which works fine on iOS 11 with SSO when its used, so its not so much of an iOS issue. Might be better to say that SSO support will be available soon (ie. when the new AppAuth version is released).

Sandboxed Safari InAppBrowser (Currently Used)
screen shot 2018-03-28 at 10 10 02

Use of Shared System Safari Browser (with AppAuth Master)
screen shot 2018-03-28 at 10 10 27

docs/auth/sso.adoc
```diff
@@ -0,0 +1,37 @@
...
+==== Configuration File in the Mobile Apps
+
+From following the link:README.adoc[AeroGear Services Auth SDK] setup, there should be a configuration file, `mobile-services.json`, inside the iOS projects. It's important that the Keycloak configuration defined in this file matches the Keycloak client that the app intends to use.
+
+'''
+=== Perform SSO
+
+Now that your two iOS apps and Keycloak clients are setup, sign into the *Email App* with the user you have previously created. You should be redirected to your device browser which should have loaded the Keycloak login page.  
+
+Next, enter in your credentials, make sure check the *Remember Me* option and login. You have now been authenticated on the *Email App* via the browser.
```
TommyJ1994: Similarly here with "Remember Me".

docs/auth/sso.adoc
```diff
@@ -0,0 +1,37 @@
...
+. Two iOS apps that have created and initialised with the Auth SDK module - link:README.adoc[AeroGear Services Auth SDK].
+
+'''
+
+=== Steps to achieve SSO via a device browser
+For this guide, let's assume that your two iOS apps are called **Email App** and **Messaging App**.
+
+==== Setup Keycloak
+
+. Ensure the "Remember Me" option is turned on. Click on *Realm Settings* section on the left hand side, and then click on *Login*. Turn on *Remember Me*.
```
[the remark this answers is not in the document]
wei-lee: are you sure? I thought without "Remember Me" the cookie will not be persistent?

docs/auth/sso.adoc
wei-lee: I disagree. Apple is recommending that `SFAuthenticationSession` should be used for all SSO, and I believe they will reject apps that uses the system browser to perform SSO. From this perspective, Apple is requiring SSO to be done using `SFAuthenticationSession`, but it is not reliable. Also sometimes it does work.

TommyJ1994: AppAuth currently uses SFAuthenticationSession. The appstore rejects seem to be due to the use of the Safari Mobile browser, rather than external user agents like Chrome, Firefox etc for Enterprise SSO use cases. AppAuth implemented Enterprise SSO use cases because of [NCCOE/NIST requirements](https://www.nccoe.nist.gov/projects/use-cases/mobile-sso).

wei-lee: Is there a difference between using Safari and Chrome or Firefox? I think what Apple's intention is that only `SFAuthenticationSession` should be used for SSO. If any other browser is used, the app will be rejected. I don't think using Chrome or Firefox will make a difference.

That's why once the new version of AppAuth is published, we can suggest if the app is for enterprise use only, it should use the external browser instead for a reliable user experience. But by default, I think it should still use `SFAuthenticationSession`.

wei-lee:
@TommyJ1994 @rachael-oregan FYI, I did a few more tests and I believe the "Remember Me" step is actually required to get SSO working on iOS.

I created 2 apps, and tried SSO with both "Remember Me" on & off. When "Remember Me" is off, I always have to re-login again in another app. But if it is enabled, I don't need to.

Issue #60, opened 2018-03-29
Update the mobile config parser to parse the version as UInt not String


rachael-oregan:
## Description
Currently the Core SDK is parsing the version property of a mobile configuration as a String. See [`MobileConfig`](https://github.com/aerogear/aerogear-ios-sdk/blob/master/modules/core/config/MobileConfig.swift#L8)  This should be refactored to `UInt` because the mobile cli tool returns the mobile configuration version as an integer. 

## Steps to reproduce
Create a client in openshift and run `mobile get clientconfig <client_id> --namespace=<namespace> -o json`. Notice the version number is and integer not a String

wtrocki:
@rachael-oregan You are hero of the day! Thanks for pointing that out.

wtrocki:
This happened to be trivial to fix.

PR #61, opened 2018-03-29
fix: Change version to integer

wtrocki:
### Motivation

Fix for https://github.com/aerogear/aerogear-ios-sdk/issues/60


wtrocki:
Note. Example app is now broken. Adding additional fix.

danielpassos:
@rachael-oregan do you wanna review before I merge it?

PR #62, opened 2018-04-02
Replace timestamp Int converter to NSString format

danielpassos:
Convert Double in Int to remove the decimal part cause an exception in some cases

> Fatal error: Double value cannot be converted to Int because the result would be greater than Int.max

Related issues: 

* [AGIOS-681](https://issues.jboss.org/browse/AGIOS-681)


wtrocki:
My vote will be to use long. Timestamp is typically an number.

danielpassos:
@wtrocki I have no problem to use long instead of String. Should I use `CLong` so? Since Swift don't have `Long`?

wtrocki:
Swift has long type but it is still int :)
`UInt64` should work with the parser and also be compatible with the Android and server side validation.

danielpassos:
@wtrocki Done, thanks!

PR #63, opened 2018-04-03
 Split ClientDeviceInformation in config and credentials 

danielpassos:
Related issues:
* [AGIOS-679](https://issues.jboss.org/browse/AGIOS-679)


modules/push/AgsPush.swift
```diff
@@ -20,19 +30,73 @@ public class AgsPush {
...
+     - parameter credentials: Credentials used to register the device
+
+     - parameter success: A block object to be executed when the registration operation finishes successfully. This block has no return value.
+
+     - parameter failure: A block object to be executed when the registration operation finishes unsuccessfully.This block has no return value and takes one argument: The `NSError` object describing the error that occurred during the registration process.
+    */
+    public func register(_ deviceToken: Data,
+                  _ config: UnifiedPushConfig,
+                  _ credentials: UnifiedPushCredentials,
+                  success: (() -> Void)!,
```
wtrocki: Remove exclamation marks

modules/push/AgsPush.swift
```diff
@@ -20,19 +30,73 @@ public class AgsPush {
...
+        let basicAuthCredentials: Data! = "\(credentials.variantID):\(credentials.variantSecret)".data(using: String.Encoding.utf8)
+        let base64Encoded = basicAuthCredentials.base64EncodedString(options: NSData.Base64EncodingOptions(rawValue: 0))
+        return ["Authorization": "Basic \(base64Encoded)"]
+    }
+
+    // Helper to transform the Data-based token into a (useful) String:
+    fileprivate func convertToString(_ deviceToken: Data?) -> String? {
+        if let token = (deviceToken as NSData?)?.description {
+            return token.replacingOccurrences(of: "<", with: "")
+                    .replacingOccurrences(of: ">", with: "")
```
wtrocki: Question: I copied that code. Is that still necessary? I tested that and there are no characters like this in token, but were not sure why this was added

wtrocki:
Have no phone. Going to verify that later

modules/push/AgsPush.swift
wtrocki: So I have added that to avoid some compat issues.

danielpassos: We can test without it and remove if still working!

modules/push/AgsPush.swift
danielpassos: I'm confused here. If I remove it I will need to do

```swift
success: @escaping (() -> Void),
failure: @escaping ((NSError) -> Void))
```

wtrocki: Yeah. We need escaping. Sorry for not being clear.

example/AeroGearSdkExampleTests/push/DeviceRegistrationTests.swift
```diff
@@ -1,89 +0,0 @@
-import AGSPush
```
wtrocki: Let's restore this tests. Removing unit tests is not the best way to go. Let's adjust them in the future.

danielpassos: Done in https://github.com/aerogear/aerogear-ios-sdk/pull/73

PR #68, opened 2018-04-04
 🚀  prepare for release

wei-lee:
### Motivation

We want to publish the modules to Cocoapods

### Description

* Remove the parent podspec file, it will not be used
* Update podspec files for each module and add license file
* Add some scripts to help with release

### Progress

- [x] Done Task


scripts/release-core.sh
```diff
@@ -0,0 +1,17 @@
...
+(cd ./modules/core && bundle exec pod lib lint)
+if [ $? -eq 0 ]; then
+  echo "Pod core is ready for release"
+else
+  echo "Pod core is not ready for release"
+  exit 1
+fi
+
+echo "Release pod core"
+(cd ./modules/core && bundle exec pod trunk push)
```
wei-lee: @wtrocki It turns out that cocoapods-release plugin doesn't work anymore. It hasn't been updated in the last 3 years. I guess that means we will use `pod trunk` instead.

@danielpassos shall we use the existing AeroGear account for publish the pods?

wei-lee:
The idea for release is that:

1. Run `pre-release.sh` to update the versions, commit them and push the tags
2. Run `release-core.sh` to publish the core module first. As it is required to be published first in order to lint the service modules.
3. Run `release-modules.sh` to publish the service modules

This is fine when everything works. However, there could be problems if there are errors when running step 2 & 3. 

I am now questioning if there are any benefits to use the same version number for all the modules. It makes it hard to manage the release process.

scripts/release-modules.sh
```diff
@@ -0,0 +1,24 @@
...
+  (cd ./modules/$module && bundle exec pod lib lint)
+  if [ $? -eq 0 ]; then
+    echo "Pod $module is ready for release"
+  else
+    echo "Pod $module is not ready for release"
+    exit 1
+  fi
+done
+
+echo "Release pods"
```
wtrocki: FYI - Let's hardcode core release before this step.

wtrocki: Edit: Got it all wrong. Idea will be to run two scripts on release. Ignore this comment.

wei-lee: I keep them separated because if there is an error during the lint step for the service modules, we will have to run this script again. But at this point, the core module is already published, and we don't need to push the core module again.

wtrocki:
> The idea for release is that:

Can we copy this to: https://github.com/aerogear/aerogear-ios-sdk/blob/master/docs/releng/release.adoc 

We could also move this out of the docs to scripts folder or other location as this is internal for this repository.

Changes loos good. We are blocked until we will get proper cocoapods account we could use with pod trunk.

wei-lee:
@wtrocki I will update the the release doc once I have verified the steps are working as expected. We don't need to create account for pushing to cocoapods, we just need an email alias to register as the owner of the pods, then we can use that to add our individual emails as owners. 

@danielpassos @matzew is there an aerogear maintainer email alias we can use?

scripts/pre-release.sh
```diff
@@ -0,0 +1,17 @@
...
+echo "Update SDK Version to $SDK_VERSION"
+sed -i.bak -E "s/DEVELOPMENT/${SDK_VERSION}/" ./modules/core/data/AgsMetadata.swift
+for module in $MODULES
+do
+   sed -i.bak -E "s/\.version([ ]*)=([ ]*)(.*)/\.version\1=\2'$SDK_VERSION'/g" ./modules/$module/*.podspec
+done
+echo "SDK version is updated to $SDK_VERSION"
+
+echo "Push release tags"
+git add -A && git commit -m "Release $SDK_VERSION"
```
wtrocki: Master could be protected. We could create PR here.

PR #71, opened 2018-04-05
fix: improve http response to return status code  (AGIOS-675)

wtrocki:
### Motivation

https://issues.jboss.org/browse/AGIOS-675

## Tasks 

- [x] Wrap http into response object.
- [x] Include status code into response object.
- [x] Extend metrics interface to use handler when metrics are published.


modules/core/http/AgsHttpResponse.swift
```diff
@@ -0,0 +1,32 @@
...
+*/
+public struct AgsHttpResponse {
+
+    /**
+     Contains response body if no error is thrown
+    */
+    public let response: Any?
+
+    /**
+     Contains error if request failed
```
wei-lee: when exactly a request is considered 'failed'? Is it when the http status code is not 200, or is it when there is an error when performing the request (like timing out, or no network etc)?

wtrocki: We are using Alamofire behind the scenes which.

https://github.com/Alamofire/Alamofire/blob/master/Documentation/Usage.md#automatic-validation

psturc:
👀

wtrocki:
@psturc that is convenience method that creates, collects metrics and publishes both. I did not exposed callbacks for the convenience method, but you could just test publish that provides callback.

PR #73, opened 2018-04-09
Read credentials from mobile-services.json

danielpassos:
Related issues:

* [AGIOS-680](https://issues.jboss.org/browse/AGIOS-680)

### DIFF
--- a/example/AeroGearSdkExample/mobile-services.json
+++ b/example/AeroGearSdkExample/mobile-services.json
@@ -23,13 +23,18 @@
       "type": "metrics",
       "url": "https://www.mocky.io/v2/5aa696133100001335e716e0",
       "config": {}
-   },
-   {
-        "id": "push",
-        "name": "push",
-        "type": "push",
-        "url": "https://ups-dm-myproject-1.193b.starter-ca-central-1.openshiftapps.com/",
-        "config": {}
-   }
+    },
+    {
+      "id": "push",
+      "name": "push",
+      "type": "push",
+      "url": "https://ups-dm-myproject-1.193b.starter-ca-central-1.openshiftapps.com/",
+      "config": {
+        "ios": {
+          "variantId": "f85015b4-a762-49a7-a36f-34a451f819a4",
+          "variantSecret": "978b35d6-7058-43b4-8c37-4dc30022ebda"
+        }
+      }
+    }
   ]
 }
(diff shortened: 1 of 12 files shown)


david-martin:
@pb82 how does this align with tasks in https://trello.com/c/eYDEQK1o/43-8-pushconfig-to-mar-sync-as-a-developer-i-want-the-configuration-generated-by-the-ups-server-to-be-available-in-the-associating

modules/push/AgsPush.swift
```diff
@@ -30,8 +31,13 @@ public class AgsPush {
         - mobileConfig: the configuration for the auth service from the service definition file
      */
     private init(_ mobileConfig: MobileService?) {
-        self.serverURL = URL(string: (mobileConfig?.url)!)!
         self.requestApi = AgsCore.instance.getHttp()
+        self.serverURL = URL(string: (mobileConfig?.url)!)!
+
+        let variant = mobileConfig?.config!["ios"]?.getObject()?["variantId"]?.getString()
```
wtrocki: 1) Is `ios` a good key? Strange that IOS is used as platform is known (for mobile app id).
@david-martin @pb82 

2) To not unwrap optionals many times we could write something like this:

```swift
let pushSecurityConfig = mobileConfig?.config!["ios"]?.getObject();
let variant = pushSecurityConfig?["variantId"]?.getString()
let secret = pushSecurityConfig??["variantSecret"]?.getString()

```

danielpassos: 1. Well, I've used the same tree of UPS standard alone.

wtrocki:
I have also noticed that unit tests were removed in previous PR
Can we restore push unit tests for registration.

example/AeroGearSdkExample/mobile-services.json
```diff
@@ -23,13 +23,18 @@
...
-   {
-        "id": "push",
-        "name": "push",
-        "type": "push",
-        "url": "https://ups-dm-myproject-1.193b.starter-ca-central-1.openshiftapps.com/",
-        "config": {}
-   }
+    },
+    {
+      "id": "push",
```
pb82: @danielpassos @wtrocki I think in the end the config generated by the mobile cli would look more like this:

```
{
  "name": "push",
  "variants": [
    {
        "uri": "https://ups-dm-myproject-1.193b.starter-ca-central-1.openshiftapps.com/",
        "type": "android",
        "googlekey": "...",
         ...
    }
  ]
}
```

matzew: does this make it harder to parse the file? since we now have no `ios` or `droid` objects ?

danielpassos: For now, all kinds of stuff should be under `config` node. There is no way to get `rootNode.variants`

PR #75, opened 2018-04-11
Add list view to Push Notification screen

wtrocki:
### Motivation

Show list view with sent push notifications the same as it happens on Android.

screen shot 2018-04-11 at 5 45 54 pm


wtrocki:
Good guy @danielpassos! Thanks

PR #77, opened 2018-04-12
fix: adjust format of the metrics to new requirements

wtrocki:
### Motivation

Adjust format of the metrics to new requirements

https://issues.jboss.org/browse/AGIOS-683

## Notes

There is no security checks implementation for IOS SDK. 
So I could not test that with the full payload.

### DIFF
--- a/modules/core/metrics/AgsMetrics.swift
+++ b/modules/core/metrics/AgsMetrics.swift
@@ -4,15 +4,21 @@ import Foundation
  AeroGear Services metrics
  */
 open class AgsMetrics: MetricsPublishable {
+    
+    private static let InitMetricsType = "init"
+    
     private let appData: AgsMetaData
     private let config: MetricsConfig
     private var publisher: MetricsPublisher!
     private var http: AgsHttp
 
+    private let defaultMetrics: [Metrics]
+
     public init(_ http: AgsHttp, _ configService: ServiceConfig) {
         self.http = http
         config = MetricsConfig(configService)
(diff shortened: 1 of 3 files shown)


wtrocki:
@psturc FYI

modules/core/metrics/AgsMetrics.swift
```diff
@@ -52,25 +58,26 @@ open class AgsMetrics: MetricsPublishable {
...
             payload[metric.identifier] = result
         }
-        publisher.publish(appendToRootMetrics(payload), handler)
     }
 
-    private func appendToRootMetrics(_ payload: MetricsData) -> [String: Any] {
+    open func publish(_ type: String, _ metrics: [Metrics], _ handler: @escaping (AgsHttpResponse?) -> Void) {
+        var data = MetricsData()
+        appendMetrics(metrics, &data)
+        appendMetrics(defaultMetrics, &data)
```
psturc: @wtrocki should we be including `defaultMetrics` in every case?

If so, [this test](https://github.com/aerogear/aerogear-ios-sdk/blob/master/example/AeroGearSdkExampleIntegrationTests/metrics/MetricsIntegrationTests.swift#L31) makes no sense now and should be removed (there would be now way how to send bad request)

This change broke both integration tests so after we resolve this, I will send a PR with the fix for the tests.

wtrocki: New payload standard assumes that metrics are always there.

PR #80, opened 2018-04-17
Docs release

wtrocki:
### Motivation

Documentation updates for release proces.


modules/core/data/AgsMetadata.swift
```diff
@@ -6,7 +6,7 @@ import Foundation
 public struct AgsMetaData {
     /** Version of Core SDK */
     // NOTE: This version is being replacted by automation script
-    public let sdkVersion = "DEVELOPMENT"
+    public let sdkVersion = "0.1.0"
```
danielpassos: This file is saying it's `0.1.0` but VERSION is saying `0.2.0` :confused:

PR #82, opened 2018-04-18
AEROGEAR-2091 - JWT validation implementation

rachael-oregan:
### Motivation

JIRA: https://issues.jboss.org/browse/AEROGEAR-2091

### Description
The [JSONWebToken](https://github.com/kreactive/JSONWebToken) library is used to verify JSON Web Tokens.  I'm using `JSONWebToken`@v2.0.0 because the latest version of the library (2.1.0) requires Swift v4.1.0 and our project uses 4.0.0.  

### Progress

- [x] `verifyClaims` function
- [x] unit tests
- [x] docs


wtrocki:
@rachael-oregan Community tend to update version rapidly. We should do that as well since we do not have official release. Feel free to bump the version as it should not cause any issues. If somehow you do not want to do it in this PR let me know. I will do that in separate one and we could rebase your PR. Have fun.

modules/auth/jwt/Jwt.swift
```diff
@@ -22,19 +16,21 @@ func base64Decode(_ input: String) -> Data? {
 
     if base64.count % 4 != 0 {
         base64.append(String(repeating: "=", count: 4 - base64.count % 4))
+    }
+    if (Data(base64Encoded: base64) == nil) {
```
wei-lee: is this `if` statement required?

modules/auth/jwt/Jwt.swift
```diff
@@ -48,20 +44,51 @@ class Jwt {
...
+            
+            let validationResult = validator.validateToken(jwt)
+            
+            if (validationResult.isValid) {
+                return true
+            } else {
+                return false
+            }
+        } catch {
+            throw Errors.generatingPublicKey("Error creating RSA public Key using the provided modulus and exponent", error)
```
wei-lee: the error message is specific about generating public key. It should either catch the specific error threw by `RSAKey.registerOrUpdateKey`,  or the catch block should be moved up after line 74.

modules/auth/jwt/Jwt.swift
```diff
@@ -48,20 +44,51 @@ class Jwt {
...
+     
+     - parameters:
+        - jwks: the JSON Web Key Set to verify the JWT against
+        - jwt: the JSON Web Token to be verified
+     
+     - throws: an `invalidModulus` error if the decoded modulus is nil, an `invalidExponent` error if the decoded exponent is nil or a `generatingPublicKey` error if the RSA public key could not be generated
+     
+     - returns: true if the JWT is valid, false otherwise
+     */
+    public static func verifyJwt(jwks: Jwks, jwt: String) throws -> Bool {
```
wei-lee: so where this function is being used? I think the `AuthService` module should be updated to actually use this method to verify the token?

modules/auth/jwt/Jwt.swift
```diff
@@ -48,20 +44,51 @@ class Jwt {
...
+        guard let modulus = base64Decode(jwks.keys[0].n) else {
```
wei-lee: so the keys in `jwks` is an array. However, depending on how many keys that are available, there might be multiple keys in the array and it's not guaranteed the first one will be a RSA type key. It should be a AES type of key.

So we should do a lookup in the array and use the first key whose type is `RSA`.

modules/auth/jwt/Jwt.swift
rachael-oregan: Ah I must have forget to remove this, thanks

modules/auth/jwt/Jwt.swift
rachael-oregan: Ya that makes sense, I'll update this to catch the `badKeyFormat` error threw by the `registerOrUpdateKey` function

modules/auth/jwt/Jwt.swift
rachael-oregan: I'll update the `AuthService` to verify the tokens

modules/auth/jwt/Jwt.swift
rachael-oregan: Ah ok,  I thought there is usually only one key returned here.  I'll update this to find the `RSA` key.  Thanks for pointing that out

rachael-oregan:
@wtrocki thanks for having a look. Ya that makes sense to bump the version.  However, to get latest version of Swift (4.1.0) I'll need to update to macOS High Sierra 10.13.4.  I can update this this evening after work cause it'll take a while.

If you have the latest version then feel free to bump the version too.  Thanks :)

rachael-oregan:
@wei-lee mind having another look please. Thanks

modules/auth/jwk/JwksManager.swift
```diff
@@ -2,6 +2,41 @@ import AGSCore
...
+ - returns: the RSA JSON Web Key or nil if not found.
+ */
+public func rsaJwk(jwks: Jwks) -> JwkContent? {
+    var rsaJwk: JwkContent?
+    for (_, jwk) in jwks.keys.enumerated() {
+        if (jwk.kty == "RSA") {
+            rsaJwk = jwk
+        }
+    }
+    return rsaJwk
```
wei-lee: I just read about how to lookup keys again, looks like we need a small improvement here:

apparently in the JWT token, there is a `kid` header, which gives you the id of the key that is used to sign/encrypt the token.

That value should be used to look up the key. So a small change to this function is required. It should take another parameter, which is the `kid` value in the JWT token, then the lookup logic should be to changed to `jwt.kid == kid`.

modules/auth/jwt/Jwt.swift
```diff
@@ -48,20 +42,54 @@ class Jwt {
...
+        }
+        guard let exponent = base64Decode(jwk.e) else {
+            throw Errors.invalidExponent("Decoded exponent cannot be nil")
         }
 
-        return JSONWebToken(header: headerValue, payload: payload!, signature: signatureValue)
+        do {
+            let publicKey: RSAKey = try RSAKey.registerOrUpdateKey(modulus: modulus, exponent: exponent, tag: "publicKey")
+              
+            let jwt = try Jwt.decode(jwt)
```
wei-lee: so this line should be moved to, which you can then access the `kid` value in the headers of the token.

Then you can use the `kid` value to lookup the right key in the `JWKS`.

rachael-oregan:
@wtrocki if you have some time would you mind reviewing these changes please?

example/AeroGearSdkExampleTests/auth/jwk/JwkManagerTests.swift
```diff
@@ -23,7 +23,7 @@ class JwkManagerTests: XCTestCase {
 
         // create JWKS
         let jwks = try? JSONSerialization.data(withJSONObject: getMockJwks(), options: [])
-        let jwksString = String(data: jwks!, encoding: String.Encoding.ascii)
+        let jwksString = String(data: jwks!, encoding: .utf8)
```
wtrocki: I will avoid `jwks!` 
We should unwrap optional.

modules/auth/AgsAuth.swift
```diff
@@ -146,15 +148,29 @@ open class AgsAuth {
...
         guard configured else {
             throw Errors.serviceNotConfigured
         }
+        
         guard let currentCredential = credentialManager.load() else {
             return nil
         }
-        guard let user = User(credential: currentCredential, clientName: keycloakConfig!.clientID) else {
+        
+        guard let jwks = jwksManager?.load(keycloakConfig!) else {
```
wtrocki: keycloakConfig! can be unwrapped above and used directly as non optional. `!` is bad as if under some circumstances it's nil app will crash

modules/auth/jwk/JwksManager.swift
```diff
@@ -80,12 +117,17 @@ class JwksManager {
                     return onCompleted!(nil, error)
                 }
             } else if let response = response.response as? [String: Any] {
+                var responseData: Data?
                 if let resData = try? JSONSerialization.data(withJSONObject: response, options: []) {
+                    responseData = resData
                     let resString = String(data: resData, encoding: .utf8)
                     self.persistJwks(keycloakConfig.realmName, resString!)
                 }
                 if onCompleted != nil {
```
wtrocki: Unwrap this instead of comparing to nil.

modules/auth/jwt/Jwt.swift
wtrocki: @rachael-oregan Was that done? Cannot see this

modules/auth/AgsAuth.swift
rachael-oregan: The `configured` var is used to do this check [here](https://github.com/rachael-oregan/aerogear-ios-sdk/blob/ff07f0e83d05a8c12702271083663f57aaf3f136/modules/auth/AgsAuth.swift#L46)

modules/auth/jwt/Jwt.swift
rachael-oregan: Yes it's been used to verify the token [here](https://github.com/rachael-oregan/aerogear-ios-sdk/blob/ff07f0e83d05a8c12702271083663f57aaf3f136/modules/auth/AgsAuth.swift#L169)

rachael-oregan:
@wtrocki I refactored to unwrap the callback optionals.  Kinda looks a bit ugly with all the `if let`s.  Let me know if this can be improved

PR #84, opened 2018-04-20
ci: releases from CircleCI

darahayes:
### Changes
* sdkVersion in `AGSMetadata.swift` is now replaced as part of `prep-release.sh`
* new `validate-release.sh` that does the following:
	* checks that `$TAG` env variable is valid
	* checks that `$TAG` is equal to the value found in `./VERSION`
	* checks each podspec file and ensures the version numbers are equal to `$TAG`
* new `release_job` in CircleCI workflow.

### Process

* Developers run `./scripts/prep-release.sh` and commit + push changes
* CircleCI notices tag
	1. build job is run first (build, test)
	2. release job is run (validate, release core, release modules)


darahayes:
@wtrocki release docs are in place. Does CircleCI already have cocoapods credentials or will we need to add these?

PR #91, opened 2018-04-23
ci: don't use bundle check

darahayes:


wtrocki:
@darahayes Could we try that out without merging to master. 
You could create tag from current branch.

I'm worried that if we start merging things that I cannot really prove they work it will be hard to get back or understand the impact.
This bundle check comes from circle ci documentation.

PR #94, opened 2018-04-27
Make metrics internal part of the core

wtrocki:
### Motivation

Make metrics internal part of the core. 
Remove protocol for end users.
Remove need for calling the sendAppAndDeviceMetrics from app.

Just leaving this here:
image


modules/core/metrics/AgsMetrics.swift
```diff
@@ -20,33 +23,29 @@ open class AgsMetrics: MetricsPublishable {
         appData = AgsCore.getMetadata()
         defaultMetrics = [AppMetrics(appData), DeviceMetrics()]
         setDefaultPublisher()
+        Timer.scheduledTimer(timeInterval: AgsMetrics.DefaultMetricsTimeout,
```
pb82: Is this to delay the sending of the metrics for a few seconds after the app started?

wtrocki:
@pb82 Thanks for taking look. I'm still trying to improve the timer part.

modules/core/metrics/AgsMetrics.swift
wtrocki: Yes. It's not ideal, but doing that on strict app startup will contribute to slow app loading times and can drive people off from using our SDK. Idea is to postpone metrics collection task so it will happen only once and in the time that views will be rendered (we are no longer can tie that to the app lifecycle). However this aproach seems little bit off. Trying to find a way to bind this to app lifecycle.

wtrocki:
There are ways to hook up into `UIApplication` object at runtime, but this seems to hacky. Delaying job using jobs API in fact slows down total execution time due to trivial work done in metrics. Used profiler for that time and found that total processing for metrics takes around 1ms, so we should not look into optimizations in that space. Networking is done in separate thread anyway.

wtrocki:
@pb82 I have added performance tests with default threshold. They should fail if we go with SDK initialization under certain level. In practice test will fail when init will take more than ~10ms

PR #95, opened 2018-05-02
AGSPush.podspec wrong filename case fixed

wojta:
### Motivation
Again there was podspec with wrong case. This would make case-senstive systems fail on `pod install`

### Description
File was renamed.


wojta:
Unfortunately it's not all, so please don't merge yet

wtrocki:
@wojta What you mean?

wtrocki:
Synced offline with @wojta. Problems were resolved. Thanks for contribution.

PR #96, opened 2018-05-03
fix: add config by type

wtrocki:
## Motivation

Use type for fetching mobile configurations instead of name.
Name is just human readable name and it's not guaranteed to be unique.

### DIFF
--- a/modules/core/AgsCore.swift
+++ b/modules/core/AgsCore.swift
@@ -26,12 +26,19 @@ public class AgsCore {
      - parameter serviceType:  service type that will be used to fetch configuration
      */
     public func getConfiguration(_ serviceType: String) -> MobileService? {
-        return config.getConfigurationByType(serviceType).first
+        let configuration = config.getConfigurationByType(serviceType)
+        if configuration.count > 1 {
+            AgsCore.logger.warning("""
+             Config contains more than one service of the same type.
+             Using configuration from the first occurence of service with that type.
+            """)
+        }
+        return configuration.first
     }
-    
+
     /**
      Get configuration for specific service reference
-     
+
      - returns: MobileService instance or nil if configuration is missing service of that type
      - parameter serviceId: id that will be used to fetch configuration
      */
--- a/modules/core/AgsCore.swift
+++ b/modules/core/AgsCore.swift
@@ -20,13 +20,23 @@ public class AgsCore {
     }
 
     /**
-     Get configuration for specific service reference
+     Get single configuration for specific service type
 
      - returns: MobileService instance or nil if configuration is missing service of that type
-     - parameter serviceRef: unique service reference uset to fetch configuration
+     - parameter serviceType:  service type that will be used to fetch configuration
      */
-    public func getConfiguration(_ serviceRef: String) -> MobileService? {
-        return config[serviceRef]
+    public func getConfiguration(_ serviceType: String) -> MobileService? {
+        return config.getConfigurationByType(serviceType).first
+    }
+    
+    /**
+     Get configuration for specific service reference
+     
+     - returns: MobileService instance or nil if configuration is missing service of that type
+     - parameter serviceId: id that will be used to fetch configuration
+     */
+    public func getConfigurationById(_ serviceId: String) -> MobileService? {
+        return config.getConfigurationById(serviceId)
     }
 
     /**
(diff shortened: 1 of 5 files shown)


modules/core/config/ServiceConfig.swift
```diff
@@ -49,16 +49,31 @@ public class ServiceConfig {
     /**
      Fetch configuration for specific type
 
+     - Parameter type: unique id of the service
```
wei-lee: this is not the `id` anymore

modules/core/config/ServiceConfig.swift
```diff
@@ -49,16 +49,31 @@ public class ServiceConfig {
...
      - return: MobileService array
      */
-    public func getConfiguration(_ serviceRef: String) -> [MobileService] {
+    public func getConfiguration(_ type: String) -> [MobileService] {
```
wei-lee: I would prefer to rename this to `getConfigurationByType`

modules/core/config/ServiceConfig.swift
```diff
@@ -49,16 +49,31 @@ public class ServiceConfig {
     /**
      Fetch configuration for specific type
 
+     - Parameter type: type of the service to fetch
      - return: MobileService array
      */
-    public func getConfiguration(_ serviceRef: String) -> [MobileService] {
+    public func getConfigurationByType(_ type: String) -> [MobileService] {
```
wei-lee: should the metrics & auth module updated as well to use this method?

wtrocki: We are using really cool operators in those SDK's. config['value'] :)

wei-lee: never mind, I see it eventually calls into `subscript`. Hmm, now we have 2 different methods to return the configurations, is it still a good idea to use `subscript`? The document is a bit unclear about how the configurations are fetched when using `subscript`

wtrocki: @wei-lee really really good point! Let's kill subscript as it's confusing now.

wtrocki:
Ready for final review.

modules/core/AgsCore.swift
```diff
@@ -20,13 +20,23 @@ public class AgsCore {
...
     /**
-     Get configuration for specific service reference
+     Get single configuration for specific service type
 
      - returns: MobileService instance or nil if configuration is missing service of that type
-     - parameter serviceRef: unique service reference uset to fetch configuration
+     - parameter serviceType:  service type that will be used to fetch configuration
+     */
+    public func getConfiguration(_ serviceType: String) -> MobileService? {
+        return config.getConfigurationByType(serviceType).first
```
aliok: @wtrocki I would log a warning statement if there is more than one config in this case: https://github.com/aerogear/aerogear-android-sdk/pull/218/files#diff-044cc3c57dde529eff9bb43b86168eebR241

josemigallas:
@wtrocki code LGTM, any verifications steps tho?

modules/core/metrics/MetricsConfig.swift
```diff
@@ -11,7 +11,7 @@ class MetricsConfig {
     public var config: MobileService?
 
     init(_ configService: ServiceConfig) {
-        if let serviceConfig = configService[sdkId] {
+        if let serviceConfig = configService.getConfigurationByType(sdkId).first {
```
aliok: similar to above, I would log a warning if there are more than one configs.

or, if possible, use AgsCore.getConfigurationByType.

aliok: or, even...
I would move the "get single config" method to `serviceConfig` and reuse that in AgsCore and here.

josemigallas:
For what it's worth, LGTM

wtrocki:
Example app should be updated but since we removing this code I haven't included get by type. If example app works it should be ok.

modules/core/config/ServiceConfig.swift
```diff
@@ -23,40 +23,31 @@ public class ServiceConfig {
...
+     - Parameter id: unique id of the service
+     - return: MobileService
      */
-    public func getConfiguration(_ serviceRef: String) -> [MobileService] {
+    public func getConfigurationById(_ id: String) -> MobileService? {
         if let config = config {
-            return config.services.filter { $0.id == serviceRef }
-        } else {
-            return []
+            return config.services.first { $0.id == id }
```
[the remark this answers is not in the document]
wtrocki: That is possible but it requires 2 unwraps vs one currently.

PR #99, opened 2018-05-03
Fix docs generator

danielpassos:


finp:
@danielpassos  
I'm getting the following:

...    gems/ruby-2.1.2/gems/jazzy-0.9.2/bin/sourcekitten: cannot execute binary file
bundler: failed to load command: jazzy (/home/pwright/.rvm/gems/ruby-2.1.2/bin/jazzy)

I'm on fedora, could we be pulling a sourcekitten binary for macos?

danielpassos:
@finp Have you run `bundle install` on the root folder before? try it?

wtrocki:
@finp This is an ruby repository. So it expects that `bundle install` was called.
Same happen for the npm (npm install is required).

To quickly fix that you can just execute `gem install jazzy`

wtrocki:
@finp you cannot do that on fedora as xcode is required.

Issue #38, opened 2018-03-07
Allow retrieve Keycloak user attributes from the SDK


wei-lee:
Keycloak allows setting arbitrary attributes on users as described [here](https://www.keycloak.org/docs/3.3/server_admin/topics/users/attributes.html).

We should update the SDK to retrieve these attributes from the user profile and allow developers to access them.

wei-lee:
@wtrocki absolutely. That's why I created those issues in Github so that people can know what kind of work they can contribute. Feel free to take the issue and work on it!

wtrocki:
Android implementation: https://github.com/aerogear/aerogear-android-sdk/pull/149

PR #101, opened 2018-05-04
Add user custom attribute retrieval 

rachael-oregan:
### Motivation

#38 

### Description

Added user custom attribute retrieval and docs


modules/auth/User.swift
```diff
@@ -238,4 +238,88 @@ public struct User {
...
+    
+    /**
+     Gets the custom user attribute of type int.
+     
+     - parameters:
+        - attributeName: the user attribute to retrieve
+     
+     - returns: the custom int attribute if it exists, otherwise nil
+     */
+    public func customIntegerAttribute(_ attributeName: String) -> Int? {
```
wei-lee: the implementation of these methods are all similar to each other, I suggest using [generics](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/Generics.html) to DRY.

rachael-oregan:
@wei-lee ya that'll look a lot cleaner, I'll refactor to use generics. Thanks

rachael-oregan:
@wei-lee I updated the implementation to use generics, let me know if I'm using it correctly. Thanks

example/AeroGearSdkExampleTests/auth/UsersTests.swift
```diff
@@ -81,4 +94,24 @@ class UsersTests: XCTestCase {
         XCTAssertFalse(userToTest!.hasRealmRole("wrongRealmRole"))
         XCTAssertFalse(userToTest!.hasClientRole(client: "clientNotExist", role: "clientRole"))
     }
+    
+    func testCustomBooleanAttribute() {
+        XCTAssertEqual(userToTest?.customAttribute("boolean") as Bool?, true)
+        XCTAssertEqual(userToTest?.customAttribute("nonExistentCustomAttribute") as Bool?, nil)
```
wei-lee: this will work, but it is not really necessary because it's already casted in the method's implementation. 
Instead, you can just declare the type when invoke `customAttribute`. So you can do:

```
var boolAttribute = userToTest?.customAttribute<Bool>("boolean");
```

It will avoid the unnecessary type cast, and also it is the recommended way of using generics.

rachael-oregan: Invoking the function this way causes the compiler to throw `Cannot explicitly specialize a generic function`. I think for it to work I need to explicitly declare the type of the variable the return value is saved in

rachael-oregan:
@wei-lee refactored to explicitly declare the type of the variable

PR #105, opened 2018-05-11
add antora structure

finp:
adding antora structure to validate publishing, will follow up with moving files to populate site


docs/package-lock.json
```diff
@@ -0,0 +1,3006 @@
+{
```
wtrocki: Can we gitignore this file?

PR #109, opened 2018-05-14
added security module with checks

StephenCoady:
### Motivation

JIRAS: [1](https://issues.jboss.org/browse/AEROGEAR-{2124}), [2](https://issues.jboss.org/browse/AEROGEAR-{2125}), [3](https://issues.jboss.org/browse/AEROGEAR-{2126})

### Description

Add individual device checks as per the JIRAs listed above.

### Progress

- [x] Jailbreak
- [x] Device lock
- [x] Debugger check 

### Verification

This PR can be verified together with [this one](https://github.com/aerogear/ios-showcase-template/pull/19).

* Clone this branch
* Use the included scripts to clone the showcase application to within the sdk.
* Check out the linked branch of the showcase app above
* Run the showcase application and verify that all device trust checks perform as expected

### DIFF
--- a/AGSSec.podspec
+++ b/AGSSec.podspec
@@ -14,8 +14,7 @@ Pod::Spec.new do |s|
                      :tag => s.version.to_s}
   s.source_files = 'modules/security/**/*.swift'
   s.frameworks   = 'UIKit', 'Foundation', 'SystemConfiguration'
-  ## Default core dependency
-  s.dependency 'AGSCore'
+  s.dependency 'DTTJailbreakDetection'
   s.requires_arc = true
 
 end
--- a/modules/security/AgsSec.swift
+++ b/modules/security/AgsSec.swift
@@ -7,16 +7,31 @@ import Foundation
  checks on the device their code is running on
  */
 public class AgsSec {
-
-    func check(_ check: SecurityCheck) -> SecurityCheckResult{
+    
+    public init(){}
+    
+    /**
+     - Perform a security check on a device.
+     
+     - Parameter check: The security check to be performed
+     - Returns: A SecurityCheckResult with a true or false property 'passed'
+     */
+    public func check(_ check: SecurityCheck) -> SecurityCheckResult{
         return check.check()
     }
 
-    func checkMany(checks: [SecurityCheck]) -> [SecurityCheckResult] {
+    /**
+     - Perform multiple security checks on a device
+     
+     - Parameter checks: The security checks to be performed
+     - Returns: An array of type SecurityCheckResult with a true or false property 'passed' for each result
+     */
+    public func checkMany(_ checks: [SecurityCheck]) -> [SecurityCheckResult] {
         var completedChecks: [SecurityCheckResult] = [];
         for value in checks {
             completedChecks.append(check(value));
         };
         return completedChecks;
     }
+
 }
--- a/modules/security/SecurityCheckResult.swift
+++ b/modules/security/SecurityCheckResult.swift
@@ -2,8 +2,8 @@ import Foundation
 
 public struct SecurityCheckResult {
 
-    let name: String
-    let passed: Bool
+    public let name: String
+    public let passed: Bool
 
     init(_ name: String, _ passed: Bool) {
         self.name = name;
--- a/modules/security/SecurityCheckResult.swift
+++ b/modules/security/SecurityCheckResult.swift
@@ -5,7 +5,7 @@ public struct SecurityCheckResult {
     public let passed: Bool
     public var result: String
 
-    init(_ name: String, _ passed: Bool, _ result: String) {
+    public init(_ name: String, _ passed: Bool, _ result: String) {
         self.name = name
         self.passed = passed
         self.result = result
--- a/modules/security/SecurityCheckResult.swift
+++ b/modules/security/SecurityCheckResult.swift
@@ -1,12 +1,13 @@
 import Foundation
 
 public struct SecurityCheckResult {
-
     public let name: String
     public let passed: Bool
+    public var result: String
 
-    init(_ name: String, _ passed: Bool) {
-        self.name = name;
-        self.passed = passed;
+    init(_ name: String, _ passed: Bool, _ result: String) {
+        self.name = name
+        self.passed = passed
+        self.result = result
     }
 }
--- a/modules/security/checks/IsDebuggerCheck.swift
+++ b/modules/security/checks/IsDebuggerCheck.swift
@@ -2,19 +2,21 @@ import Foundation
 
 public class IsDebuggerCheck: SecurityCheck {
     public var name = "Debugger"
-  
-    public init(){}
-    
+    private let passing = "Debugger not detected"
+    private let failing = "Debugger detected"
+
+    public init() {}
+
     /**
      - Check if the device is running in Debug mode.
-     
+
      - Returns: A Security Check result with a true or false passing property
      */
     public func check() -> SecurityCheckResult {
         #if DEBUG
-        return SecurityCheckResult(self.name, false)
+            return SecurityCheckResult(self.name, false, self.failing)
         #else
-        return SecurityCheckResult(self.name, true)
+            return SecurityCheckResult(self.name, true, self.passing)
         #endif
     }
 }
--- a/modules/security/checks/IsDebuggerCheck.swift
+++ b/modules/security/checks/IsDebuggerCheck.swift
@@ -1,19 +1,20 @@
 import Foundation
 
-public class IsEmulatorCheck: SecurityCheck {
-    public let name = "Emulator"
-
+public class IsDebuggerCheck: SecurityCheck {
+    public var name = "Debugger"
+  
     public init(){}
     
     /**
-     - Check if the device is running in an emulator.
+     - Check if the device is running in Debug mode.
      
      - Returns: A Security Check result with a true or false passing property
      */
     public func check() -> SecurityCheckResult {
-        #if (arch(i386) || arch(x86_64) && os(iOS))
+        #if DEBUG
         return SecurityCheckResult(self.name, false)
-        #endif
+        #else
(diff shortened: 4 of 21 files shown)


modules/security/checks/IsJailbrokenCheck.swift
```diff
@@ -0,0 +1,19 @@
+import Foundation
+import DTTJailbreakDetection
+
+public class IsJailbrokenCheck: SecurityCheck {
+    var name = "Is Jailbroken Check"
```
aidenkeating: Could we have these as a `let` in the check implementations so they're constant?

Also, slightly unrelated, do we need our `SecurityCheck` protocol to enforce `get` and `set` of the name property? I think it can just be `get` as it shouldn't change, right?

modules/security/checks/IsJailbrokenCheck.swift
```diff
@@ -0,0 +1,19 @@
...
+    
+    /**
+     - Check if the device is Jailbroken.
+     
+     - Returns: A Security Check result with a true or false passing property
+     */
+    internal func check() -> SecurityCheckResult {
```
aidenkeating: Out of curiousity? Why internal here?

wtrocki:
@StephenCoady Totally forgot about this one. 
Showcase app should be working now.

modules/security/checks/IsJailbrokenCheck.swift
StephenCoady: Yeah I've already changed this locally, it needs to be public. thanks

StephenCoady:
@aidenkeating I will ping you as soon as I have the other PR open so verification will be easier. @wtrocki thanks, just writing the code now. thanks guys

StephenCoady:
ping @wtrocki @aidenkeating @wtrocki @rachael-oregan could you guys take another look if you get a chance please

modules/security/AgsSec.swift
```diff
@@ -7,16 +7,32 @@ import Foundation
...
+     */
+    public func checkMany(_ checks: [SecurityCheck]) -> [SecurityCheckResult] {
         var completedChecks: [SecurityCheckResult] = [];
         for value in checks {
             completedChecks.append(check(value));
         };
         return completedChecks;
     }
+
+    public init(){
```
rachael-oregan: Should we make this the first method of the class?

modules/security/checks/IsDeviceLockCheck.swift
```diff
@@ -0,0 +1,23 @@
+import Foundation
+import LocalAuthentication
+
+public class IsDeviceLockCheck: SecurityCheck {
+    public var name = "Device Lock"
+    public var description = "This check will detect if a lock screen has been set on the device. Not setting a lock screen on the device can make it vulnerable to unauthorised access."
```
aidenkeating: In the showcase app we're performing a check to see if the iOS version is iOS9 or later (https://github.com/StephenCoady/ios-showcase-template/blob/fa29b8190a60439f3b1b82bdac7e6a49592bb57f/ios-showcase-template/services/DevicetrustService.swift#L39). Should we document that here? Does this check only work on iOS 9 or later?

aidenkeating: Is is possible to have these properties as `let` so they're constants?

modules/security/SecurityCheckResult.swift
```diff
@@ -2,11 +2,13 @@ import Foundation
 
 public struct SecurityCheckResult {
 
-    let name: String
-    let passed: Bool
+    public let name: String
+    public let passed: Bool
+    public let description: String
```
rachael-oregan: I'm not sure if this is necessary?

modules/security/checks/IsDebuggerCheck.swift
```diff
@@ -0,0 +1,20 @@
+import Foundation
+
+public class IsDebuggerCheck: SecurityCheck {
+    public var name = "Debugger"
+    public var description = "This check will detect if the app is running in a debug mode. A debuggable app allows calls within the app to be traced."
+    
+    public init(){}
```
aidenkeating: Do we need to explicitly specify an `init` function?

wtrocki:
I really wanted to do some unit testing here but nature of this changes make this really hard.

wtrocki:
@StephenCoady `./scripts/development.sh` - execute this

modules/security/SecurityCheckResult.swift
rachael-oregan: I'm just thinking of consistency really.  I don't think we give a description in the other SDKs.  No harm in adding it but I think we should try keep it consistent across the SDKs anyway

modules/security/checks/IsDeviceLockCheck.swift
wtrocki: @aidenkeating That is not important. IOS 8 is prehistory.

wtrocki: This will crash if someone will downgrade the target in order to test things on old phone. What I'm saying is that we do not need to care that much as IOS8 as target is not existing and there will be no production apps with this system.

modules/security/SecurityCheckResult.swift
StephenCoady: The reason I added it is because there is a view in the iphone app which makes use of it. I have no issues removing that view altogether. Its a popup modal here: https://github.com/aerogear/ios-showcase-template/pull/19/files#diff-3f381704ad8cb017d4bf7eef0423a8e1R94

modules/security/checks/IsDebuggerCheck.swift
aidenkeating: Ah, it's not public, ok cool! Thanks 👍

modules/security/checks/IsDeviceLockCheck.swift
StephenCoady: Actually the way swift works these need to be a var because they have a getter in the SecurityCheck protocol and "let declarations cannot be computed properties"

aidenkeating: Ah didn't pay attention to the protocol, makes sense, thanks!

StephenCoady: ignore that ^ I thought you meant doing this on the protocol

aidenkeating:
@StephenCoady Just as a side note, the clone script seems to have a syntax error

```
(cd showcase && && LOCAL=true pod install)
```

aidenkeating:
@StephenCoady There's an issue with device lock message being red when the device lock is enabled and green when it's disabled. Other than that everything looks good

AGSSec.podspec
```diff
@@ -14,8 +14,7 @@ Pod::Spec.new do |s|
                      :tag => s.version.to_s}
   s.source_files = 'modules/security/**/*.swift'
   s.frameworks   = 'UIKit', 'Foundation', 'SystemConfiguration'
-  ## Default core dependency
-  s.dependency 'AGSCore'
+  s.dependency 'DTTJailbreakDetection'
```
wei-lee: I think the security module will still depend on the AGSCore module, for the metrics stuff.

modules/security/AgsSec.swift
```diff
@@ -7,16 +7,31 @@ import Foundation
  checks on the device their code is running on
  */
 public class AgsSec {
-
-    func check(_ check: SecurityCheck) -> SecurityCheckResult{
+    
+    public init(){}
```
wei-lee: I think the constructor of the all the service modules should be the same, and it should take the configurations parameter. It will allow us to add backend services quite easily if we want to.

I think a `ServiceModule` interface would help to ensure all the service modules are using at least 1 same constructors.

@wtrocki WDYT?

modules/security/AgsSec.swift
```diff
@@ -7,16 +7,31 @@ import Foundation
  checks on the device their code is running on
  */
 public class AgsSec {
```
wei-lee: I also think it's much better to rename the module to `AGSSecurity`, I don't see the benefit of using abbreviation here.

modules/security/AgsSec.swift
StephenCoady: With regard to the email thread going around last week, what if there is no configuration needed, as is the case for security at the moment? Are we saying it can just be empty then?

modules/security/AgsSec.swift
wei-lee: I think it's very common to use `Auth` for `Authentication`, but I am not sure if there is a convention to use `Sec` for `Security`. Maybe I am overthinking. Anyway, let's not worry about it for now. We can deal with this later.

modules/security/AgsSec.swift
wei-lee: If all the service modules implement the same interface, then it should take at least the configuration parameter. But the service module itself can choose using or not using the configuration object. It makes things easier if the module need to use the configuration in the future, especially for developers who consume the modules, as it means adding the configurations won't be a breaking changes of the SDK.

Anyway, it should be a separate discussion, it should not block this PR. You can continue as it is for now, we can make the changes later if we decide to do so.

modules/security/AgsSec.swift
wtrocki: Had the same feeling. Let's change that.

TommyJ1994:
Should there be more consistent naming across the SDK's, at least with the positive/negative scheme?

_Android_
- `NOT_ROOTED`
- `NOT_IN_EMULATOR`
- `SCREEN_LOCK_ENABLED`
- `NO_DEBUGGER`

_iOS_
- `IsJailbrokenCheck` 
- `IsEmulatorCheck`
- `IsDeviceLockCheck`
- `IsDebuggerCheck`

StephenCoady:
@TommyJ1994 you are right. I have pushed with the refactor in place. Could you take a look and let me know if thats ok.

TommyJ1994:
@StephenCoady Looks good!

PR #117, opened 2018-05-22
update antora structure

laurafitzgerald:
### Motivation
Update the antora structure to match other SDK docs format/syntax

### Progress

- [x] Move docs under /docs/modules/ROOT/pages
- [x] Remove all other folders under modules
- [x] update README to use new routes
- [x] small adoc fixes

PR #122, opened 2018-05-29
feat: Security Check Metrics AEROGEAR-2779

StephenCoady:
### Motivation

This commit adds security check metrics publishing ability to the security module

JIRA: [https://issues.jboss.org/browse/AEROGEAR-2779]

### Description

Added new publish methods, changed current MetricsData to be of type any instead of a dictionary as security metrics are slightly different to other metrics on the server side currently. 

### Verification steps
This accompanying [PR](https://github.com/aerogear/ios-showcase-template/pull/23) can be used for verification purposes.

* Install mobile core locally
* From the mobile tab, create a metrics service
* Point the app at this metrics service
* Run the showcase application and navigate to the Device trust page
* Metrics should be populated in Grafana
* Change the DeviceTrustService.swift file to the following: https://gist.github.com/StephenCoady/b4d96ec4630decdc17d968bdc0fa3c65 and run the app again to publish all metrics at once as opposed to one at a time.


modules/core/metrics/engine/MetricsPublisher.swift
```diff
@@ -2,17 +2,23 @@ import Foundation
 
 /**
    Represents single metrics data
+   We use Any as type to be able to use Array and Dictionary
  */
-public typealias MetricsData = [String: Any]
+public typealias MetricsData =  Any
+
+/**
+   Root object that is being send to server to collect metrics
```
wtrocki: remove `to collect metrics` - I do not know what I meant by this!

modules/core/metrics/engine/MetricsPublisher.swift
```diff
@@ -2,17 +2,23 @@ import Foundation
...
+ */
+public typealias MetricsRoot = [String: Any]
 
 /**
    Protocol for classes that can publish or store metrics payload
  */
 public protocol MetricsPublisher {
-    /**
+    /**ß
```
wtrocki: :)

modules/security/AgsSecurity.swift
```diff
@@ -8,6 +9,11 @@ import Foundation
  */
 public class AgsSecurity {
     
+    private let serviceId = "security"
+    
+    /** instance of the Security SDK */
```
wtrocki: of `the Core SDK`

modules/security/AgsSecurity.swift
```diff
@@ -8,6 +9,11 @@ import Foundation
...
+    public let core = AgsCore.instance;
```
wtrocki: -100 for this. If we do it this way we can have nasty issues when this class will be loaded before instance is created. We should use that in directly as in other SDK and do not assign  value on the field level.

modules/security/AgsSecurity.swift
```diff
@@ -36,4 +62,25 @@ public class AgsSecurity {
...
+    /**
+     - Perform multiple security checks on a device and also publish them to the metrics service
+     
+     - Parameter checks: The security checks to be performed
+     - Returns: An array of type SecurityCheckResult with a true or false property 'passed' for each result
+     */
+    public func checkManyAndPublishMetric(_ checks: [SecurityCheck]) -> [SecurityCheckResult] {
+        let results = checkMany(checks)
+        let securityResults = SecurityCheckResultMetric(results)
+        core.getMetrics().publish(serviceId, [securityResults], { (response: AgsHttpResponse?) -> Void in
```
wtrocki: Noticed bug here. When there is no metrics configuration we are not going to call handlers.
See publish method implementation:
```
 guard let activePublisher = publisher else {
 
 }
```

modules/security/deviceTrust/SecurityCheckResult.swift
```diff
@@ -5,7 +5,7 @@ import Foundation
  
  A successful check means that the environment the application is running in is more secure and no action is required, as opposed to signalling if a certain feature is enabled.
  */
-public struct SecurityCheckResult {
+public struct SecurityCheckResult: Encodable {
```
wtrocki: Verify if this still needs to be encodable. 
We have been remapping that to `Metrics` anyway.

modules/security/metrics/SecurityCheckResultMetric.swift
```diff
@@ -0,0 +1,39 @@
+import Foundation
+import AGSCore
+/**
+ Protocol for the metric representation of a `SecurityCheckResult`.
```
wtrocki: It's not protocol

modules/security/metrics/SecurityCheckResultMetric.swift
```diff
@@ -0,0 +1,39 @@
...
+        self.results = results;
+    }
+    
+    public func collect() -> MetricsData {
+        var collectArray : [Any] = [];
+        for result in self.results {
+            do {
+                let singleCheck = try result.adaptToDictionary()
+                let mappedObj = [
+                    "name": singleCheck["name"]!,
```
wtrocki: -100 for `!` Use `??` `singleCheck["name"] ?? ""`. Parser will also handle optionals as empty so we do not need that anyway.

tests/AeroGearSdkExample.xcodeproj/project.pbxproj
```diff
@@ -788,11 +798,11 @@
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CODE_SIGN_STYLE = Automatic;
-				DEVELOPMENT_TEAM = GHPBX39444;
+				DEVELOPMENT_TEAM = VBNQKKFDW6;
 				INFOPLIST_FILE = AeroGearSdkExample/Info.plist;
 				IPHONEOS_DEPLOYMENT_TARGET = 9.0;
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks";
-				PRODUCT_BUNDLE_IDENTIFIER = org.aerogear.AeroGearSdkExample;
+				PRODUCT_BUNDLE_IDENTIFIER = org.scoady.aerogear;
```
wtrocki: Revert this change

tests/AeroGearSdkExample.xcodeproj/project.pbxproj
```diff
@@ -805,11 +815,11 @@
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CODE_SIGN_STYLE = Automatic;
-				DEVELOPMENT_TEAM = GHPBX39444;
+				DEVELOPMENT_TEAM = VBNQKKFDW6;
```
wtrocki: Revert this change

tests/AeroGearSdkExample.xcodeproj/project.pbxproj
```diff
@@ -805,11 +815,11 @@
...
 				INFOPLIST_FILE = AeroGearSdkExample/Info.plist;
 				IPHONEOS_DEPLOYMENT_TARGET = 9.0;
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks";
-				PRODUCT_BUNDLE_IDENTIFIER = org.aerogear.AeroGearSdkExample;
+				PRODUCT_BUNDLE_IDENTIFIER = org.scoady.aerogear;
```
wtrocki: Revert this change

wtrocki:
@StephenCoady This looks really good. Left some minor comments. Once this is done I will verify that with the running metrics server.

docs/security/README.adoc
```diff
@@ -84,4 +84,34 @@ To invoke multiple checks the `AGSSecurity.checkMany` function is used. This ret
 ----
 let checks = [IsDeviceLockCheck(), IsEmulatorCheck()]
 let results = security.checkMany(checks)
+----
```
laurafitzgerald: @StephenCoady there are some [changes](https://github.com/aerogear/aerogear-ios-sdk/pull/117) to the structure of the documentation that will affect where this docs go. Depending on which pr gets merged first we may need to sync around updating the location of this file.

StephenCoady: @laurafitzgerald merge race 😜

yeah thats fine though. yours will probably land first so I will just rebase and convert this doc. I did the antora stuff for cordova so I am familiar.

modules/security/AgsSecurity.swift
```diff
@@ -22,7 +25,27 @@ public class AgsSecurity {
...
+     
+     - Parameter check: The security check to be performed
+     - Returns: A SecurityCheckResult with a true or false property 'passed'
+    */
+    public func checkAndPublishMetric(_ check: SecurityCheck) -> SecurityCheckResult {
+        let result = check.check();
+        AgsCore.instance.getMetrics().publish(serviceId, [SecurityCheckResultMetric([result])], { (response: AgsHttpResponse?) -> Void in
+            if let error = response?.error {
+                AgsCore.logger.error("An error has occurred when sending check metrics: \(error)")
+                return
```
darahayes: I don't know how important it is, but one of the acceptance criteria in this [trello card](https://trello.com/c/pYV8kinn/50-3-as-a-developer-if-metrics-server-is-not-provisioned-available-and-i-use-security-metrics-in-my-application-my-app-android-ios) says:

> The developer should be able to handle the scenario when the Metrics Service is not provisioned or not available.

Just wondering how a developer using our SDK could handle the case where the metrics were not published successfully. Perhaps they would want to execute some special behaviour in the event that the metrics were not published.

This implementation would mean that there's no way to handle this case in code. The only way a developer would know is by reading logs.

darahayes: You can dismiss this comment :)

darahayes:
I realised I never commented. I've verified that this works!

Issue #124, opened 2018-05-31
ignore the cases when lookup service configurations based on types


wei-lee:
## Expected Behavior

When lookup the service configurations based types, the case shouldn't really matter. E.g. I can use "keycloak" or "Keycloak"

By the look of code, I think this is a bug in the ios sdk, but I could be wrong.

PR #125, opened 2018-06-05
AEROGEAR-2905 - Ignore cases when looking up configurations based on …

rachael-oregan:
…types

## Motivation

JIRA: https://issues.jboss.org/browse/AEROGEAR-2905

GitHub Issue: #124 

## Description

Refactored service configuration *type* lookup to ignore the letter case requested and search for lowercase 

## Progress

- [x] refactor service config *type* lookup to ignore the letter case used in the request

### DIFF
--- a/modules/core/config/ServiceConfig.swift
+++ b/modules/core/config/ServiceConfig.swift
@@ -30,7 +30,7 @@ public class ServiceConfig {
      */
     public func getConfigurationByType(_ type: String) -> [MobileService] {
         if let config = config {
-            return config.services.filter { $0.type == type }
+            return config.services.filter { $0.type.lowercased() == type.lowercased() }
         } else {
             return []
         }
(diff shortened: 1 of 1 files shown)


modules/core/config/ServiceConfig.swift
```diff
@@ -30,7 +30,7 @@ public class ServiceConfig {
      */
     public func getConfigurationByType(_ type: String) -> [MobileService] {
         if let config = config {
-            return config.services.filter { $0.type == type }
+            return config.services.filter { $0.type.lowercased() == type.lowercased() }
```
wtrocki: Hint on good programming. `type.lowercased()` can be done outside the filter.

wei-lee:
@wtrocki can you explain why you dislike this? Feel free to reply on the mailing list and it is still open.

wtrocki:
@wei-lee IMHO Any error with configuration should be fixed in the source = APB descriptions. Ignoring case it's just one of example of things that can get wrong. For example someone changed type from `push` to `"Unified Push Server"` 
This change is trivial so ok to merge, but wanted to state that this is not going to resolve problem of the APB types being changed constantly :)

TL;DR - I do not want to argue over something so trivial.

wei-lee:
@wtrocki I don't really think the cases should be treated as an error, but others should be.

PR #126, opened 2018-06-06
release version 0.4.0

StephenCoady:


modules/core/data/AgsMetadata.swift
```diff
@@ -6,7 +6,7 @@ import Foundation
 public struct AgsMetaData {
     /** Version of Core SDK */
     // NOTE: This version is being replacted by automation script
-    public let sdkVersion = "0.3.1"
+    public let sdkVersion = "0.4.0
```
wtrocki: Let's terminate this!

PR #128, opened 2018-06-08
 🔖 update version number to 1.0.0-alpha

wei-lee:
### Motivation

Prepare for release 1.0.0-alpha

### Description

Update version numbers

### Progress

- [x] Done Task


wei-lee:
@wtrocki @darahayes mind reviewing this?

wei-lee:
@darahayes done

wtrocki:
I have run podspec tests and I think it fails to recognize the qualifier 1.0.0-anything for core. We may need to specify that explicity to avoid using old version.
I will need 10 minutes to check that

PR #130, opened 2018-06-12
fix: Application is crashing when push config is missing

wtrocki:
### Motivation

Fix application crashing when push configuration is missing. This is even more critical than Auth as push is being initialized on start making application unusable.

### DIFF
--- a/modules/push/AgsPush.swift
+++ b/modules/push/AgsPush.swift
@@ -39,7 +39,7 @@ public class AgsPush {
     init(_ mobileConfig: MobileService?, _ requestApi: AgsHttpRequestProtocol) {
         self.requestApi = requestApi
         guard let config = mobileConfig else{
-           AgsCore.logger.info("Missing configuration for UPS. Push registration will be disabled");
+           AgsCore.logger.warning("Missing configuration for UPS. Push registration will be disabled");
            return;
         }
         self.serverURL = URL(string: (config.url))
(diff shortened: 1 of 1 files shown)


wtrocki:
Basic rule of thumb: We should never use `!` in swift especially when handling json that can change format.

modules/push/AgsPush.swift
```diff
@@ -38,13 +38,17 @@ public class AgsPush {
      */
     init(_ mobileConfig: MobileService?, _ requestApi: AgsHttpRequestProtocol) {
         self.requestApi = requestApi
-        self.serverURL = URL(string: (mobileConfig?.url)!)!
-
-        let pushSecurityConfig = mobileConfig?.config!["ios"]?.getObject()
+        guard let config = mobileConfig else{
+           AgsCore.logger.info("Missing configuration for UPS. Push registration will be disabled");
```
wei-lee: I think warning/error here is more appropriate. If this is used, apparently the developer want to use push, but there is no configuration, so it should at least a warning.

modules/push/AgsPush.swift
```diff
@@ -63,9 +67,15 @@ public class AgsPush {
                          _ config: UnifiedPushConfig,
                          success: @escaping (() -> Void),
                          failure: @escaping ((Error) -> Void)) {
+        
+        guard let url = self.serverURL, let credentials = self.credentials else {
+            AgsCore.logger.info("Cannot register to UPS. Missing configuration");
```
wei-lee: similar here should be warning/error as well.

PR #131, opened 2018-06-19
AEROGEAR-3063 Expose push configuration errors to user

aidenkeating:
### Motivation
 It would be nice to have an error exposed to a developer if the service configuration isn't defined. This is done for Auth and it allows users to react to the config not being there.

### Description

Expose a `noServiceConfigurationFound` error to the end user if the configuration for the push service is not defined in `mobile-services.json`.

Also, fix a typo the clone showcase script.


aidenkeating:
@wei-lee Would you mind taking a look? This would require a release to CocoaPods, but I think it makes sense, considering there's no nice way to determine whether the configuration is found or not at the moment.

modules/push/AgsPush.swift
```diff
@@ -38,17 +44,7 @@ public class AgsPush {
...
-           AgsCore.logger.warning("Missing configuration for UPS. Push registration will be disabled");
-           return;
-        }
-        self.serverURL = URL(string: (config.url))
-        let pushSecurityConfig = mobileConfig?.config?["ios"]?.getObject()
-        let variant = pushSecurityConfig?["variantId"]?.getString()
-        let secret = pushSecurityConfig?["variantSecret"]?.getString()
-        if let variant = variant, let secret = secret {
-           self.credentials = UnifiedPushCredentials(variant, secret)
-        }
```
wei-lee: we can probably leave them here. But I would suggest creating a new variable for the class called "pushConfig".

This variable will be initialised in the constructor. If the config is missing, it will be null.

Then in every method, the first thing is to check if the `pushConfig` is set. If it is not set, an exception should be thrown.

aidenkeating:
@wei-lee I *think* I addressed your feedback

modules/push/AgsPush.swift
```diff
@@ -66,35 +77,35 @@ public class AgsPush {
...
+            self.requestApi.post(registerUrl, body: postData, headers: headers, { response in
+                guard let requestError = response.error else {
+                    self.saveClientDeviceInformation(deviceToken, pushConfig.url)
+                    success()
+                    return
+                }
+                failure(requestError)
+            })
+        } else {
+            throw Errors.noServiceConfigurationFound
```
wei-lee: this is ok. But I think use `guard` on L82 instead of `if` and throw the error earlier will make it easier to read the code.

aidenkeating: I forgot about `guard`, will do! Thanks 👍

aidenkeating:
@wei-lee Updated to use `guard`

PR #132, opened 2018-06-21
release version 1.0.0-alpha.2

aidenkeating:
### Motivation

Release `1.0.0-alpha.2`. Naming based on https://semver.org/#spec-item-9. Not sure if we want to do `1.0.0-alpha.1` instead, it seems confusing to me.

@wei-lee Does this look right to you? Following https://github.com/aerogear/aerogear-ios-sdk/blob/master/docs/contrib/release.adoc


aidenkeating:
@wei-lee Mind taking a look?

wei-lee:
@aidenkeating I think we can use `alpha.1`?

wei-lee:
@danielpassos Not sure I would agree with that. But it probably doesn't matter anyway. @aidenkeating which one you prefer? We do not need to align the SDK versions at the moment.

aidenkeating:
@wei-lee @danielpassos According to semver `1.0.0-alpha < 1.0.0-alpha.1` so I think we should go with `1.0.0-alpha.1` (I do find it confusing though) https://semver.org/#spec-item-11. I'll make a new release branch now.

aidenkeating:
Replaced with https://github.com/aerogear/aerogear-ios-sdk/pull/133

PR #135, opened 2018-06-27
Prepare for release 1.0.0-alpha.2

wei-lee:
### Motivation

Prepare for release 1.0.0-alpha.2

ping @aidenkeating


modules/core/data/AgsMetadata.swift
```diff
@@ -6,7 +6,7 @@ import Foundation
 public struct AgsMetaData {
     /** Version of Core SDK */
     // NOTE: This version is being replacted by automation script
-    public let sdkVersion = "1.0.0-alpha.1"
+    public let sdkVersion = "1.0.0-alpha.2-alpha.1"
```
aidenkeating: This looks to not have been replaced properly?

wei-lee: @aidenkeating good catch. Will fix it.

wei-lee:
@aidenkeating should be fixed now.

PR #137, opened 2018-07-03
remove user facing docs

pb82:
### Motivation

Remove user facing docs. All user facing docs should be in docs.aerogear.org. Only dev / technical documentation should remain in the repos.


wtrocki:
@pb82 There are also some scripts in `./scripts` folder that will fail after we remove folder. In particular:

- https://github.com/aerogear/aerogear-ios-sdk/blob/master/scripts/generate-service.sh
Let's remove docs folder creation here

- https://github.com/aerogear/aerogear-ios-sdk/blob/master/scripts/docs.sh
Let's generate docs to api.

PR #140, opened 2018-07-12
Sync SDK initial wrapper

wtrocki:
### Motivation

Add Sync SDK initial wrapper

### DIFF
--- a/modules/security/AgsSecurity.swift
+++ b/modules/security/AgsSecurity.swift
@@ -11,7 +11,6 @@ public class AgsSecurity {
 
     private let serviceId = "security"
     
-    public static let instance: AgsSecurity = AgsSecurity()
 
     /**
      The initialise method of AgsSecurity
--- a/modules/security/AgsSecurity.swift
+++ b/modules/security/AgsSecurity.swift
@@ -1,5 +1,5 @@
-import Foundation
 import AGSCore
+import Foundation
 
 /**
  AeroGear Services Security SDK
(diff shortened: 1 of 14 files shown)


modules/sync/AgsSync.swift
```diff
@@ -0,0 +1,34 @@
...
+        let config = AgsCore.instance.getConfiguration(serviceName)
+        let httpInterface = AgsCore.instance.getHttp()
+        return AgsSync(config)
+    }()
+   
+    public static let instance = SyncService()
+ 
+    public let client: ApolloClient
+
+    init(_ mobileConfig: MobileService?) {
```
danielpassos: Can we make the init/constructor receives what they really need instead of the whole config? [I'm trying this approach on the Android](https://github.com/aerogear/aerogear-android-sdk/pull/261/files#diff-92898add42c5be3e84d96a6af0b7c834R22) to make easy to create services out of the box like in unit tests.

wtrocki: I will need to refactor every other SDK and this is not that simple. 
`MobileService` it's config for individual service.

josemigallas: I agree this is open for extension.
However, the name is confusing. What about `syncConfig`?

modules/security/AgsSecurity.swift
```diff
@@ -8,9 +8,11 @@ import AGSCore
  checks on the device their code is running on
  */
 public class AgsSecurity {
-    
+
     private let serviceId = "security"
     
+    public static let instance: AgsSecurity = AgsSecurity()
```
josemigallas: Does this belong to this PR?

PR #141, opened 2018-07-12
Add new appauth external browser support

wtrocki:
### Motivation

During the login SSO not always work immediately. 
Doing login in safari is usually recommended. 
I have added external browser support in order to perform more mobile friendly flow where login is happening in safari.

### DIFF
--- a/modules/auth/authenticator/OIDCAuthenticator.swift
+++ b/modules/auth/authenticator/OIDCAuthenticator.swift
@@ -82,8 +82,8 @@ public class OIDCAuthenticator: Authenticator {
         - error: the error returned in the `callback` function.  Will be nil if the token exchange was successful
      */
     func startAuthorizationFlow(byPresenting: OIDAuthorizationRequest, presenting: UIViewController, callback: @escaping (_ credentials: OIDCCredentials?, _ error: Error?) -> Void) -> OIDAuthorizationFlowSession {
-        
-        let flowCallback : OIDAuthStateAuthorizationCallback = { authState, error in
+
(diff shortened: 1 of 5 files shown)


modules/auth/authenticator/OIDCAuthenticator.swift
```diff
@@ -92,7 +93,18 @@ public class OIDCAuthenticator: Authenticator {
...
+        if(authConfig.useExternalUserAgent){
+            let externalUserAgent = OIDExternalUserAgentIOSCustomBrowser.customBrowserSafari();
+            return OIDAuthState.authState(byPresenting: byPresenting,
+                                          externalUserAgent: externalUserAgent,
+                                          callback:flowCallback)
+        }else {
+             return OIDAuthState.authState(byPresenting: byPresenting,
+                                           presenting: presenting,
+                                           callback: flowCallback)
+        }
```
wei-lee: indentation seems to be off here?

modules/auth/config/AuthenticationConfig.swift
```diff
@@ -18,8 +20,9 @@ public struct AuthenticationConfig {
         - redirectURL: the redirect URL for the developers app
         - minTimeBetweenJwksRequests: The minimum time, in minutes, between Json web key set requests. Default value is 1400 (1 day)
```
wei-lee: should update the docs as well?

wtrocki: Yes. Thanks for pointing that out.

wtrocki:
Thanks for reviews. Rebased changes and formatted code using our swiftlint profile.

PR #142, opened 2018-08-01
Remove metrics service on the README services list

danielpassos:


wtrocki:
What's the motivation here?

danielpassos:
@wtrocki There is no Metrics _module_ anymore. Metrics service was moved to Core _module_

PR #143, opened 2018-08-14
AEROGEAR-7744 - Rename Security Checks to Device Checks

ziccardi:
Changed the Debug Check to check if a debugger is attached

### Motivation

JIRA: https://issues.jboss.org/browse/AEROGEAR-7744

### Description

Changed the Security checks to Device checks.
Reimplemented the debug check so that it verifies if a debugger is attached or not


danielpassos:
I have reviews this using the [ios-showcase-template](https://github.com/aerogear/ios-showcase-template/tree/AEROGEAR-7744) and some checks made me a little confused.

screenshot 2018-08-21 13 41 54

ziccardi:
@danielpassos This PR changes both the name of the checks (from security check to Device Check) and their meaning, removing the double negatives.
For that reason, there is no "NoJailbreakCheck' anymore, but there is now a 'IsJailbroken' check.
The NoJailbrakCheck was returning `true` when the jailbreak was **NOT** enabled, while the new one returns `true` when the jailbreak **IS** enabled. The same is for Emulator and Debugger. That is the reason why you see them the opposite way in the showcase.
To make the showcase work, I need to update its code to reflect the new logic, but to update the showcase, I first need to release the SDK.

danielpassos:
@ziccardi Gotcha. Feel free to use the [ios-showcase-template#AEROGEAR-7744](https://github.com/aerogear/ios-showcase-template/tree/AEROGEAR-7744). I already have replace the checks there.

PR #146, opened 2018-08-29
Add subscription transport

wtrocki:
### Motivation

Add support for subscriptions.
Related to https://github.com/aerogear/aerogear-ios-example-apps/pull/8


wtrocki:
@danielpassos This PR adding subscription support but there is still some issue in upstream implementation vs our server implementation as messages are not delivered back despite being connected.

MixedNetworkTransport.swift
```diff
@@ -0,0 +1,8 @@
+//
+//  MixedNetworkTransport.swift
+//  AGSAuth
+//
+//  Created by Wojciech Trocki on 8/28/18.
+//
+
+import Foundation
```
danielpassos: I'm totally confused about this file. What is this about?

wtrocki: So... Apollo IOS do not support websockets out of the box. You need to use something like MixedNetworkTransport that will allow you do choice between subscriptions and classical http. 
We need custom one since we need to provide token as headers. This is not documented but you can find transport implementations in latest release of the SDK.

danielpassos:
I have tested it using the [aerogear-ios-example-apps#AEROGEAR-7873](https://github.com/aerogear/aerogear-ios-example-apps/tree/AEROGEAR-7873) how fix the meme schema problem. This branch was created in top o https://github.com/aerogear/aerogear-ios-example-apps/pull/8

danielpassos:
@wtrocki Should we fix the build or ignore it for now?

wtrocki:
@danielpassos Did you got messages delivered back to the app?

MixedNetworkTransport.swift
wtrocki: Ah.. This file is just xcode mess. Removing it.

danielpassos:
Yes, it works like a charm. That is why I have approved it

wtrocki:
WOW - that's strange. It looks like it did not work for me because of different reasons.

PR #148, opened 2018-09-26
Implemented token refresh using performAction

ziccardi:
### Motivation

JIRA: [https://issues.jboss.org/browse/AEROGEAR-7853]

### Description

A new `currentUser` method has been added accepting the `autoRefresh` parameter.
If the access token needs to be refreshed, it gets refreshed automatically.

### DIFF
--- a/modules/auth/AuthHeaderProvider.swift
+++ b/modules/auth/AuthHeaderProvider.swift
@@ -17,25 +17,6 @@ public class AuthHeaderProvider: HeaderProvider {
 
     /**
      Gets the header to be used to perform HTTP requests.
-     If the accessToken needs to be updated, this method will synchronously make the request, blocking until the result is received.
-     */
-    public func getHeaders() -> [String: String] {
-        var result: [String: String]? = nil
-        let group = DispatchGroup()
-        group.enter()
-        
-        getHeaders() { headers in
-            result = headers
-            group.leave()
-        }
-        
-        group.wait()
-        
-        return result!
-    }
-
-    /**
-     Gets the header to be used to perform HTTP requests.
      If needed, the `accessToken` will be automatically refreshed and the callback will be always called with a fresh token
      */
     public func getHeaders(completionHandler: @escaping ([String: String]) -> Void ) -> Void {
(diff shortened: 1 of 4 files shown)


wtrocki:
@ziccardi would it be possible to also update memeolist app that can be used to verify this functionality?

modules/auth/AuthHeaderProvider.swift
```diff
@@ -15,15 +15,39 @@ public class AuthHeaderProvider: HeaderProvider {
         self.auth = auth
     }
 
+    /**
+     Gets the header to be used to perform HTTP requests.
+     If the accessToken needs to be updated, this method will synchronously make the request, blocking until the result is received.
+     */
     public func getHeaders() -> [String: String] {
+        var result: [String: String]? = nil
+        let group = DispatchGroup()
```
wtrocki: If this request will be slow (and it could be because of network latency) it will lag entire application ui. We could just build async chain instead and change top level api.
I believe that we are free to do it as this was still not documented feature.

ziccardi: Yeah, however the `performActionWithFreshTokens` is always executed in the main thread, so I don't think we would get much benefit:

https://github.com/openid/AppAuth-iOS/blob/dd945e14c4b5b889eeb712361cf4476868b87217/Source/OIDAuthState.h#L219

ziccardi:
@wtrocki : updated the memeolist in [MEMEO#17](https://github.com/aerogear/aerogear-ios-example-apps/pull/17)
I needed some time because I was getting a deadlock into the memeolist and I had to remove all the sync calls to fix that

modules/sync/SyncNetworkTransport.swift
```diff
@@ -79,6 +79,17 @@ public class SyncNetworkTransport: NetworkTransport {
         self.sendOperationIdentifiers = sendOperationIdentifiers
     }
 
+    private func getHeaders() -> Promise<[String:String]> {
+        
+        let promise = Promise<[String:String]> { (fullfill, error) in
```
wtrocki: [Optional] Using promise here makes no sense. Get headers method itself allow to pass handler and this will work in all callers.

